PR #3, opened 2016-11-19
Support bwm stop processing new jobs when a file changes

iwiznia:
@flodnv please review. /cc @coleaeason 

Needed for https://github.com/Expensify/Expensify/issues/40404

Web and Salt PRs incoming. We can merge this anyway, since no one will be using this version of the lib.

Tests:
- Invoke bwm without the versionWatchFile param. Check it works.
- Invoke bwm with the versionWatchFile param:
  - Check it works if the file doesn't exist.
  - Check that it stops processing new jobs when the file time changes.
  - Check that it waits till all processing jobs finish before dying.
- Bonus feature: You can now stop bwm on dev with CTRL+C 😄

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -13,19 +13,17 @@ use Expensify\Bedrock\Exceptions\Jobs\RetryableException;
  * 3. Spawn a worker for that job
  * 4. Goto 1
  * After N cycle in the loop, we exit
+ * If the versionWatchFile modified time changes, we stop processing new jobs and exit after finishing all running jobs.
  *
- * Usage: `Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration>]`
+ * Usage: `Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration> --versionWatchFile=<file>]`
  */
-
-// We need to handle signals with callbacks
-declare (ticks = 1);
 
 // Verify it's being started correctly
 if (php_sapi_name() !== "cli") {
     throw new Exception('This script is cli only');
 }
 
-$options = getopt('', ['host::', 'port::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::']);
+$options = getopt('', ['host::', 'port::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::']);
 $jobName = isset($options['jobName']) ? $options['jobName'] : null;
 $maxLoad = isset($options['maxLoad']) && floatval($options['maxLoad']) ? floatval($options['maxLoad']) : 0;
 $maxLoopIteration = isset($options['maxIterations']) && intval($options['maxIterations']) ? intval($options['maxIterations']) : 0;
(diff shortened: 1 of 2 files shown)


iwiznia:
Please, remind me to create the tag when we merge this.

bin/BedrockWorkerManager.php
```diff
@@ -118,7 +106,13 @@
 
         // Get any job managed by the BedrockWorkerManager
         $response = null;
-        while (!$response && !$receivedSIGINT) {
+        while (!$response) {
+            clearstatcache(true, $versionWatchFile);
```
quinthar: Can you add a comment here explaining what's going on?

bin/BedrockWorkerManager.php
```diff
@@ -118,7 +106,13 @@
...
+            $newVersionWatchFileTimestamp = $versionWatchFile && file_exists($versionWatchFile) ? filemtime($versionWatchFile) : false;
+            if ($versionWatchFile && $newVersionWatchFileTimestamp !== $versionWatchFileTimestamp) {
+                $logger->info('Version watch file changed, stop processing new jobs');
+                break 2;
```
quinthar: Whoa, wtf, can `break` take a parameter?  :mindblown:  What does the parameter do?

flodnv: Haha, [per the doc](http://php.net/manual/en/control-structures.break.php) it "tells it how many nested enclosing structures are to be broken out of. The default value is 1, only the immediate enclosing structure is broken out of."

quinthar: Wow, I had no idea.  But given that I had no idea, I'd suggest we not use this behavior -- where possible, let's aim for the simplest possible solution that is the most widely understood.  Is there another solution that someone as junior as me would understand automatically?

flodnv: Yeah not the biggest fan of this weird construct either. It seems to me like any code using this thing can be refactored into using functions instead (so that `return` can be used and it will be very apparent where we're "breaking" from). That being said, not sure how easy/hard/how much sense it makes in this existing script

quinthar: Well, replacing the `while (true)` with a `while ($keepLooping)` could work.

bin/BedrockWorkerManager.php
iwiznia: Added

bin/BedrockWorkerManager.php
iwiznia: Replacing `while(true)` with `while ($keepLooping)` is not enough. I'd need to add the condition to the other while and a couple of extra ifs to skip the following statements. 
I agree with the sentiment of the comment, but to make this code actually look better a better refactor is needed (break things up into functions and/or create a class to hold all the logic).
I can do that, if we think is valuable, but I think is outside the scope of the change.

quinthar:
What about just including a comment on the break statement?

On Nov 22, 2016 2:33 PM, "Ionatan Wiznia" <notifications@github.com> wrote:

*@iwiznia* commented on this pull request.
------------------------------

In bin/BedrockWorkerManager.php
<https://github.com/Expensify/Bedrock-PHP/pull/3>:

> @@ -118,7 +106,13 @@

         // Get any job managed by the BedrockWorkerManager
         $response = null;
-        while (!$response && !$receivedSIGINT) {
+        while (!$response) {
+            clearstatcache(true, $versionWatchFile);
+            $newVersionWatchFileTimestamp = $versionWatchFile &&
file_exists($versionWatchFile) ? filemtime($versionWatchFile) : false;
+            if ($versionWatchFile && $newVersionWatchFileTimestamp
!== $versionWatchFileTimestamp) {
+                $logger->info('Version watch file changed, stop
processing new jobs');
+                break 2;

Replacing while(true) with while ($keepLooping) is not enough. I'd need to
add the condition to the other while and a couple of extra ifs to skip the
following statements.
I agree with the sentiment of the comment, but to make this code actually
look better a better refactor is needed (break things up into functions
and/or create a class to hold all the logic).
I can do that, if we think is valuable, but I think is outside the scope of
the change.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/Expensify/Bedrock-PHP/pull/3>, or mute the thread

iwiznia:
> What about just including a comment on the break statement?

Sure, adding that.

iwiznia:
Updated, retested.

flodnv:
I'm 👍 , how about you @quinthar?

flodnv:
@iwiznia create the tag when we merge this.

https://github.com/Expensify/Bedrock-PHP/pull/3#issuecomment-261676229

iwiznia:
Thanks for the reminder, created!

PR #5, opened 2017-01-26
Add an optional repeat parameter to udpateJob()

tgolen:
I'm not sure what the procedure is here, sooooooooooo

### DIFF
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -142,16 +142,18 @@ class Jobs extends Plugin
      *
      * @param int   $jobID
      * @param array $data
+     * @param string $repeat      (optional) see https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.md#repeat-syntax
      *
      * @return array
      */
-    public function updateJob($jobID, $data)
+    public function updateJob($jobID, $data, $repeat = null)
     {
         return $this->call(
             "UpdateJob",
             [
                 "jobID"    => $jobID,
                 "data"     => $data,
+                "repeat"   => $repeat,
             ]
         );
     }


tgolen:
Also, the matching bedrock PR has already been merged 

https://github.com/Expensify/Bedrock/pull/95

src/Jobs.php
```diff
@@ -142,16 +142,18 @@ public function getJob($name, $timeout = 0)
...
      */
-    public function updateJob($jobID, $data)
+    public function updateJob($jobID, $data, $repeat = null)
     {
         return $this->call(
             "UpdateJob",
             [
                 "jobID"    => $jobID,
                 "data"     => $data,
+                "repeat"   => $repeat,
```
quinthar: Does this work if `$repeat` is null?  Wouldn't it send a blank repeat and trigger a `402 Malformed repeat`?

tgolen: I think null is OK. It's used for all the optional parameters for all the other methods.

tgolen: Especially `createJob()` which has the same bedrock validation for the `repeat` param.

quinthar: Oh, cool.

quinthar:
Ok I've merged it, but I'm not sure how we're deploying this.  I think there is a composer package somewhere, but I don't know what the next step is.  @righdforsa What's next here?

tgolen:
I think we need to create a tag, and then update `composer.json` in the web repo to point to the new tag.

iwiznia:
Oh, shoot... Yeah, you needed to run the `package.sh` script of this repo, which would change the version number and create a tag. Once you do that, you need to change the version number in composer.json of web-e and run `../script/composer.sh update expensify/php-libs`

PR #7, opened 2017-01-30
Allow setting of connection header for queueJob

timszot:
@iwiznia Please review these changes.

Added additional params to `queueJob` and `createJob` to allow setting the `'Connection'` header.

Reference: https://github.com/Expensify/Web-Expensify/pull/16198#discussion_r98462964

### Tests
Tested with changes in https://github.com/Expensify/Web-Expensify/pull/16198
1. Uploaded Square receipt into dev.
2. After parsing completed, changed the category from Uncategorized to a category that has a mapped MCC in the policy settings.
3. Confirmed www-prod/UpdateMerchantToMCCUsage was queued in the jobs table and returned `202` code success code.

### DIFF
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -38,6 +38,20 @@ class Jobs extends Plugin
     const STATE_QUEUED = "QUEUED";
 
     /**
+     * "Connection" header option to wait for a response.
+     *
+     * @var string
+     */
+    const CONNECTION_WAIT = "wait";
+
+    /**
+     * "Connection" header option to forget and not wait for a response.
+     *
+     * @var string
+     */
+    const CONNECTION_FORGET = "forget";
+
+    /**
      * Calls the Jobs plugin.
      *
      * @param string $method  Method to call
@@ -49,6 +63,8 @@ class Jobs extends Plugin
      * @throws MalformedAttribute
      * @throws SqlFailed
      * @throws GenericError
+     *
+     * @return array
      */
     public function call($method, $headers = [], $body = '')
     {
@@ -79,6 +95,7 @@ class Jobs extends Plugin
             throw new SqlFailed("SQL failed for job $job: {$codeLine}");
         }
 
+        // 202 code is a successful job creation using the "Connection: forget" header
         if (!in_array($responseCode, [200, 202])) {
             throw new GenericError("Generic error for job $job");
         }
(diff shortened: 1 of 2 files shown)


src/Jobs.php
```diff
@@ -79,7 +79,7 @@ public function call($method, $headers = [], $body = '')
             throw new SqlFailed("SQL failed for job $job: {$codeLine}");
         }
 
-        if ($responseCode !== 200) {
+        if (!in_array($responseCode, [200, 202])) {
```
iwiznia: Can you add a comment saying 202 means `Successfully queued`?

src/Jobs.php
```diff
@@ -89,30 +89,32 @@ public function call($method, $headers = [], $body = '')
...
      *
-     * @param string $name
-     * @param array  $data        (optional)
-     * @param string $firstRun    (optional)
-     * @param string $repeat      (optional) see https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.md#repeat-syntax
-     * @param bool   $unique      Do we want only one job with this name to exist?
-     * @param int    $priority    (optional) Specifiy a job priority. Jobs with higher priorities will be run first.
-     * @param int    $parentJobID (optional) Specify this job's parent job.
+     * @param string      $name
+     * @param array|null  $data        (optional)
```
iwiznia: Damn, were these `|null` necessary? I thought they weren't when the param default is already `null`.

src/Jobs.php
```diff
@@ -265,22 +267,24 @@ public function queryJob($jobID)
...
-     * @param bool   $unique   Do we want only one job with this name to exist?
-     * @param int    $priority Specifiy a job priority. Jobs with higher priorities will be run first.
+     * @param string      $name
+     * @param array|null  $data        (optional)
+     * @param string|null $firstRun    (optional)
+     * @param string|null $repeat      (optional) see https://github.com/Expensify/Bedrock/blob/master/plugins/Jobs.md#repeat-syntax
+     * @param bool        $unique      Do we want only one job with this name to exist?
+     * @param int         $priority    (optional) Specify a job priority. Jobs with higher priorities will be run first.
+     * @param int|null    $parentJobID (optional) Specify this job's parent job.
+     * @param string|null $connection  (optional) Specify 'Connection' header.
```
iwiznia: Thought about making this a boolean? Like `$connectionForget = false`

src/Jobs.php
timszot: They would be when Phan is concerned, I've fixed enough of these 😄

src/Jobs.php
timszot: The connection header can have different values than just `forget` though, so I thought this was a better approach, no preference to change it though.

iwiznia: I think we only have `forget` and `wait` and `wait` is default.
Anyway, I'd say either make this a boolean or add `forget` and `wait` as constants of this class.

iwiznia:
After all changes are done and approved, you need to increase the version and create the tag (using package.sh in this repo).

timszot:
@iwiznia 
Updated code and commented back.

timszot:
I copied the changes made here to my Web-Expensify vendor directory to test all of this. Followed logs and confirmed it was a `Fire and forget` situation. All looked great.

iwiznia:
Awesome! Then you only need to create the version and I'll merge.

PR #8, opened 2017-02-17
Catch throwable

iwiznia:
This will allow us to catch php errors that go uncatched now, like [this one](https://www.expensify.com/_devportal/tools/logSearch/#query=request_id:("vqgsba")+AND+timestamp:[2017-02-01T14:41:16.679Z+TO+2017-02-01T16:41:16.679Z]&index=logs_2017-02-01)

Tests:
In BWM we log the exception as a param, since the LoggerInterface doesn't have `->exception` as a method like `Log` does.
```
$caca = null;
try {
    $caca->caca();
} catch (Throwable $e) {
    Log::alert('Yay!', ['error' => $e]);
}
```
Log:
```
Feb 17 15:50:46 vagrant-ubuntu-trusty-64 php-cgi: wmNorj /caca.php ionatan@expensify.com !web! ?api? [alrt] Yay! ~~ error: 'Error: Call to a member function caca() on null in /vagrant/Web-Expensify/caca.php:7#012Stack trace:#012#0 {main}'
```

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -206,7 +206,7 @@ try {
                                 'extraParams' => $extraParams,
                             ]);
                             $bedrock->jobs->retryJob($job['jobID'], $e->getDelay(), $worker->getData());
-                        } catch (Exception $e) {
+                        } catch (Throwable $e) {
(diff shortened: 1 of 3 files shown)


flodnv:
Oops, we should update README.md

PR #12, opened 2017-03-06
Add writeConsistency as a command line param

iwiznia:
Issue: https://github.com/Expensify/Expensify/issues/47084

We added `writeConsistency` as a config option but didn't added it to the command line options.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -15,7 +15,7 @@ use Expensify\Bedrock\Exceptions\Jobs\RetryableException;
  * After N cycle in the loop, we exit
  * If the versionWatchFile modified time changes, we stop processing new jobs and exit after finishing all running jobs.
  *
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -15,15 +15,15 @@
  * After N cycle in the loop, we exit
  * If the versionWatchFile modified time changes, we stop processing new jobs and exit after finishing all running jobs.
  *
- * Usage: `Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration> --versionWatchFile=<file>]`
+ * Usage: `Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration> --versionWatchFile=<file>] --writeConsistency=<consistency>`
```
pecanoro: After looking at the code, I assumed this parameter is optional so this comment is wrong and confused me initially 😄

iwiznia:
Updated. Also, forgot to create the new version before, done it already.

PR #13, opened 2017-03-09
Prevent leak

iwiznia:
By having a circular reference between `Client` and its plugins, PHP can't free the objects, thus leaking sockets. This is not a big problem for web requests, since they are short lived, but it is for long running processes that instantiate many `Client` objects.

/cc @cead22

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -2,6 +2,7 @@
 
 use Expensify\Bedrock\Client;
 use Expensify\Bedrock\Exceptions\Jobs\RetryableException;
+use Expensify\Bedrock\Jobs;
 
 /*
  * BedrockWorkerManager
@@ -82,6 +83,7 @@ try {
 
     // Connect to Bedrock -- it'll reconnect if necessary
     $bedrock = new Client();
+    $jobs = new Jobs($bedrock);
 
     // Begin the infinite loop
     $loopIteration = 0;
@@ -122,7 +124,7 @@ try {
             }
             try {
                 // Attempt to get a job
-                $response = $bedrock->jobs->getJob($jobName, 60 * 1000); // Wait up to 60s
+                $response = $jobs->getJob($jobName, 60 * 1000); // Wait up to 60s
             } catch (Exception $e) {
                 // Try again in 60 seconds
                 $logger->info('Problem getting job, retrying in 60s', ['message' => $e->getMessage()]);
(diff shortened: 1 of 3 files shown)


iwiznia:
Integration test is failing due to urban airship api failing, not even sure what urban airship is or what the fuck it has to do with auto export. I think you are safe to ignore.

iwiznia:
Oh, some kind of push notifications that should be disabled in tests.

bin/BedrockWorkerManager.php
```diff
@@ -82,6 +83,7 @@
 
     // Connect to Bedrock -- it'll reconnect if necessary
     $bedrock = new Client();
+    $jobs = new Jobs($bedrock);
```
cead22: If you knew how much I fought Mike to standardize on this way of using BR (which I believe also makes it easier to test)...

quinthar: Can you clarify how this design makes it easier to test?  What tests do we have in place that are using this new design?

quinthar:
Aha, actually I see what you are saying -- yes the garbage collector will correctly destruct objects on script shutdown, even with circular references.  However, objects *without* circular references are actually destructed *before* then: the moment they are `unset()`, as shown here:

```
Davids-MacBook-Pro-2:Downloads dbarrett$ cat circular.php 
<?php
class ChildClass {
    public function __construct($parent) {
        print "ChildClass constructor\n";
        //$this->parent = $parent;
    }

    public function __destruct() {
        print "ChildClass destructor\n";
    }
}

class ParentClass {
    public function __construct() {
        print "ParentClass constructor\n";
        $this->child = new ChildClass($this);
    }

    public function __destruct() {
        print "ParentClass destructor\n";
    }
}

print "Start\n";
$test = new ParentClass();
print "Created\n";
unset($test);
print "End\n";

Davids-MacBook-Pro-2:Downloads dbarrett$ php circular.php 
Start
ParentClass constructor
ChildClass constructor
Created
ParentClass destructor
ChildClass destructor
End
Davids-MacBook-Pro-2:Downloads dbarrett$ 
```

That's really interesting, I didn't know PHP had that behavior.  One more reason to dislike garbage collection and appreciate explicit memory management ;)

bin/BedrockWorkerManager.php
quinthar: Actually ignore this comment; doesn't really matter as the original design just doesn't work in light of PHP not freeing circular dependencies until script termination, which kinda blows the whole idea up.

iwiznia:
Yep, the problem is not that they don't get deallocated ever, is that they don't get deallocated when they go unused (instead they get deallocated when the program ends). I also tried doing explicit `unset`s but that didn't work so I ended up with the solution you see here,

PR #14, opened 2017-03-24
[HOLD] Refactor bwm into a class

iwiznia:
@flodnv @pecanoro can you please review?

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -1,8 +1,7 @@
 <?php
 
+use Expensify\Bedrock\BedrockWorkerManager;
 use Expensify\Bedrock\Client;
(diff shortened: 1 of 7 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -59,196 +72,9 @@
...
-    // Connect to Bedrock -- it'll reconnect if necessary
-    $bedrock = new Client();
-    $jobs = new Jobs($bedrock);
-
-    // Begin the infinite loop
-    $loopIteration = 0;
-    while (true) {
-        if ($loopIteration === $maxLoopIteration) {
-            $logger->info("We did all our loops iteration, shutting down");
-            exit(0);
```
iwiznia: I think this was a bug and might have created some stuck jobs. We are not waiting for the children to finish processing. Changed it in the new implementation.

PR #15, opened 2017-04-12
Add simple demo framework for Bedrock::Jobs; add requestID; add retries

quinthar:
This does the following:

1. Adds a simple script that demonstrates Bedrock::Jobs with a trivial worker.
2. Includes a few minor internal cleanups, with comments on each inside the diff.
3. Provides a new API to fix https://github.com/Expensify/Expensify/issues/50199 (see https://github.com/Expensify/Web-Expensify/pull/17025)
4. Fixes https://github.com/Expensify/Expensify/issues/50137

Sorry for globbing multiple issues and cleanups into a single PR; I wanted to use the demo harness I built to solve those other issues.

Assigning to @iwiznia for review this and https://github.com/Expensify/Web-Expensify/pull/17025

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -71,34 +71,37 @@ if ($maxLoad <= 0) {
 Client::configure($bedrockConfig);
 
 $logger = Client::getLogger();
+$logger->info('Starting BedrockWorkerManager', ['maxLoopIteration' => $maxLoopIteration]);
+
 $stats = Client::getStats();
 $versionWatchFileTimestamp = $versionWatchFile && file_exists($versionWatchFile) ? filemtime($versionWatchFile) : false;
 
+// Wrap everything in a general exception handler so we can handle error
+// conditions as gracefully as possible.
 try {
-    $logger->info('Starting BedrockWorkerManager', ['maxLoopIteration' => $maxLoopIteration]);
-
-    if (!file_exists('/proc/loadavg')) {
-        throw new Exception('are you in a chroot?  If so, please make sure /proc is mounted correctly');
-    }
-
     // Connect to Bedrock -- it'll reconnect if necessary
     $bedrock = new Client();
     $jobs = new Jobs($bedrock);
 
-    // Begin the infinite loop
+    // Loop a finite number of times and then self destruct.  This is to guard
+    // against memory leaks, as we assume there is some other script that will
+    // restart this when it dies.
     $loopIteration = 0;
     while (true) {
+        // Is it time to self destruct?
         if ($loopIteration === $maxLoopIteration) {
             $logger->info("We did all our loops iteration, shutting down");
-            exit(0);
+            break;
         }
-
         $loopIteration++;
         $logger->info("Loop iteration", ['iteration' => $loopIteration]);
 
         // Step One wait for resources to free up
         while (true) {
             // Get the latest load
+            if (!file_exists('/proc/loadavg')) {
+                throw new Exception('are you in a chroot?  If so, please make sure /proc is mounted correctly');
+            }
             $load = sys_getloadavg()[0];
             if ($load < $maxLoad) {
                 $logger->info('Load is under max, checking for more work.', ['load' => $load, 'MAX_LOAD' => $maxLoad]);
@@ -109,12 +112,17 @@ try {
             }
         }
 
-        // Get any job managed by the BedrockWorkerManager
+        // Poll the server until we successfully get a job
         $response = null;
         while (!$response) {
-            // php's filemtime results are cached, so we need to clear that cache or we'll be getting a stale modified time.
+            // Watch a version file that will cause us to automatically shut
+            // down if it changes.  This enables triggering a restart if new
+            // PHP is deployed.
+            //
+            // Note: php's filemtime results are cached, so we need to clear
+            //       that cache or we'll be getting a stale modified time.
             clearstatcache(true, $versionWatchFile);
-            $newVersionWatchFileTimestamp = $versionWatchFile && file_exists($versionWatchFile) ? filemtime($versionWatchFile) : false;
+            $newVersionWatchFileTimestamp = ($versionWatchFile && file_exists($versionWatchFile)) ? filemtime($versionWatchFile) : false;
             if ($versionWatchFile && $newVersionWatchFileTimestamp !== $versionWatchFileTimestamp) {
                 $logger->info('Version watch file changed, stop processing new jobs');
 
@@ -122,8 +130,10 @@ try {
                 // just wait for child processes to finish.
                 break 2;
             }
+
+            // Query the server for a job
             try {
-                // Attempt to get a job
+                // Call GetJob
                 $response = $jobs->getJob($jobName, 60 * 1000); // Wait up to 60s
             } catch (Exception $e) {
                 // Try again in 60 seconds
@@ -166,11 +176,15 @@ try {
             $logger->info("Looking for worker '$workerFilename'");
             if (file_exists($workerFilename)) {
                 // The file seems to exist -- fork it so we can run it.
+                //
+                // Note: By explicitly ignoring SIGCHLD we tell the kernel to
+                //       "reap" finished child processes automatically, rather
+                //       than creating "zombie" processes.  (We don't care
+                //       about the child's exit code, so we have no use for the
+                //       zombie process.)
                 $logger->info("Forking and running a worker.", [
                     'workerFileName' => $workerFilename,
                 ]);
-                // Ignore SIGCHLD signal, which should help 'reap' zombie processes, forcing zombies to kill themselves
-                // in the event that the parent process dies before the child/zombie)
                 pcntl_signal(SIGCHLD, SIG_IGN);
                 $pid = pcntl_fork();
                 if ($pid == -1) {
@@ -245,10 +259,13 @@ try {
         }
     }
 
-    // We wait for all children to finish before dying.
-    $status = null;
-    pcntl_wait($status);
 } catch (Exception $e) {
     $message = $e->getMessage();
     $logger->alert('BedrockWorkerManager.php exited abnormally', ['exception' => $e]);
 }
+
+// We wait for all children to finish before dying.
+$logger->info('Stopping BedrockWorkerManager, waiting for children');
+$status = null;
+pcntl_wait($status);
+$logger->info('Stopped BedrockWorkerManager');
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -176,14 +176,24 @@ try {
                     $errorMessage = pcntl_strerror(pcntl_get_last_error());
                     throw new Exception("Unable to fork because '$errorMessage', aborting.");
                 } elseif ($pid == 0) {
-                    // **NOTE: There is a crazy hack where we depend on the
-                    //         info below being sent exactly as currently
-                    //         worded; do not make any changes
+                    // If we are using a global REQUEST_ID, reset it to indicate this is a new process.
                     $logger->info("Fork succeeded, child process, running job", [
                         'name' => $job['name'],
                         'id' => $job['jobID'],
                         'extraParams' => $extraParams,
                     ]);
+                    if (isset($GLOBALS['REQUEST_ID'])) {
+                        // Reset the REQUEST_ID and re-log the line so we see
+                        // it when searching for either the parent and child
+                        // REQUEST_IDs.
+                        mt_srand(getmypid());
+                        $GLOBALS['REQUEST_ID'] = substr(base64_encode(sha1(mt_rand())),0,6); // random 6 character ID
+                        $logger->info("Fork succeeded, child process, running job", [
+                            'name' => $job['name'],
+                            'id' => $job['jobID'],
+                            'extraParams' => $extraParams,
+                        ]);
+                    }
                     $stats->counter('bedrockJob.create.'.$job['name']);
 
                     // Include the worker now (not in the parent thread) such
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -143,7 +143,7 @@ try {
             // it like:
             //
             //     $worker = new $workerName( $job );
-            //     $worker->safeRun( );
+            //     $worker->run( );
             //
             // The optional path info allows for jobs to be scheduled
             // selectively.  For example, you may have separate jobs scheduled
@@ -176,12 +176,14 @@ try {
                     $errorMessage = pcntl_strerror(pcntl_get_last_error());
                     throw new Exception("Unable to fork because '$errorMessage', aborting.");
                 } elseif ($pid == 0) {
+                    // **NOTE: There is a crazy hack where we depend on the
+                    //         info below being sent exactly as currently
+                    //         worded; do not make any changes
                     $logger->info("Fork succeeded, child process, running job", [
                         'name' => $job['name'],
                         'id' => $job['jobID'],
                         'extraParams' => $extraParams,
                     ]);
-
                     $stats->counter('bedrockJob.create.'.$job['name']);
 
                     // Include the worker now (not in the parent thread) such
@@ -221,6 +223,7 @@ try {
                         }
                     });
 
+                    // The forked worker process is all done.
                     $stats->counter('bedrockJob.finish.'.$job['name']);
                     exit(1);
                 } else {
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -16,21 +16,32 @@ use Expensify\Bedrock\Jobs;
  * After N cycle in the loop, we exit
  * If the versionWatchFile modified time changes, we stop processing new jobs and exit after finishing all running jobs.
  *
- * Usage: `Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration> --versionWatchFile=<file> --writeConsistency=<consistency>]`
+ * Usage: `Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<iteration> --versionWatchFile=<file> --writeConsistency=<consistency>]`
  */
 
 // Verify it's being started correctly
 if (php_sapi_name() !== "cli") {
+    // Throw an exception rather than just output because we assume this is
+    // being executed on a webserver, so no STDOUT.  Hopefully they've
+    // configured a general uncaught exception handler, and this will trigger
+    // that.
     throw new Exception('This script is cli only');
 }
 
+// Parse the command line and verify the required settings are provided
 $options = getopt('', ['host::', 'port::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::']);
-$jobName = isset($options['jobName']) ? $options['jobName'] : null;
-$maxLoad = isset($options['maxLoad']) && floatval($options['maxLoad']) ? floatval($options['maxLoad']) : 0;
-$maxLoopIteration = isset($options['maxIterations']) && intval($options['maxIterations']) ? intval($options['maxIterations']) : 0;
-if (!$maxLoopIteration) {
-    $maxLoopIteration = 1000;
+$workerPath = @$options['workerPath'];
+if (!$workerPath) {
+    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency>]\r\n";
+    exit(1);
 }
+
+// Add defaults
+$jobName = @$options['jobName'] ?: '*'; // Process all jobs by default
+$maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
+$maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
+
+// ?? Why do we do this? 
 $bedrockConfig = [];
 if (isset($options['host'])) {
     $bedrockConfig['host'] = $options['host'];
@@ -59,42 +70,44 @@ if (isset($options['failoverPort'])) {
 if (isset($options['writeConsistency'])) {
     $bedrockConfig['writeConsistency'] = $options['writeConsistency'];
 }
-$versionWatchFile = isset($options['versionWatchFile']) ? $options['versionWatchFile'] : null;
-$workerPath = isset($options['workerPath']) ? $options['workerPath'] : null;
-if (!$jobName || !$maxLoad || !$workerPath) {
-    throw new Exception('Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration> --writeConsistency=<consistency>]');
-}
-if ($maxLoad <= 0) {
-    throw new Exception('Maximum load must be greater than zero');
-}
-
 Client::configure($bedrockConfig);
 
+// Prepare to use the host logger, if configured
 $logger = Client::getLogger();
-$logger->info('Starting BedrockWorkerManager', ['maxLoopIteration' => $maxLoopIteration]);
+$logger->info('Starting BedrockWorkerManager', ['maxIterations' => $maxIterations]);
 
-$stats = Client::getStats();
+// If --versionWatch is enabled, begin watching a version file for changes
+$versionWatchFile = @$options['versionWatchFile'];
 $versionWatchFileTimestamp = $versionWatchFile && file_exists($versionWatchFile) ? filemtime($versionWatchFile) : false;
 
 // Wrap everything in a general exception handler so we can handle error
 // conditions as gracefully as possible.
+$stats = Client::getStats();
 try {
+    // Validate details now that we have exception handling
+    if (!is_dir($workerPath)) {
+        throw new Exception("Invalid --workerPath path '$workerPath'");
+    }
+    if ($maxLoad <= 0) {
+        throw new Exception('--maxLoad must be greater than zero');
+    }
+
     // Connect to Bedrock -- it'll reconnect if necessary
     $bedrock = new Client();
     $jobs = new Jobs($bedrock);
 
-    // Loop a finite number of times and then self destruct.  This is to guard
-    // against memory leaks, as we assume there is some other script that will
-    // restart this when it dies.
-    $loopIteration = 0;
+    // If --maxIterations is set, loop a finite number of times and then self
+    // destruct.  This is to guard against memory leaks, as we assume there is
+    // some other script that will restart this when it dies.
+    $iteration = 0;
     while (true) {
         // Is it time to self destruct?
-        if ($loopIteration === $maxLoopIteration) {
+        if ($maxIterations > 0 && $iteration >= $maxIterations) {
             $logger->info("We did all our loops iteration, shutting down");
             break;
         }
-        $loopIteration++;
-        $logger->info("Loop iteration", ['iteration' => $loopIteration]);
+        $iteration++;
+        $logger->info("Loop iteration", ['iteration' => $iteration]);
 
         // Step One wait for resources to free up
         while (true) {
@@ -142,29 +155,28 @@ try {
             }
         }
 
-        if (!$response) {
-            $logger->info('Got no response from bedrock... Continuing.');
-            continue;
-        }
-
         // Found a job
         if ($response['code'] == 200) {
-            // BWM jobs are '/' separated names, the last component of which indicates the name of the worker to
-            // instantiate to execute this job:
-            // arbitrary/optional/path/to/workerName
+            // BWM jobs are '/' separated names, the last component of which
+            // indicates the name of the worker to instantiate to execute this
+            // job:
+            //
+            //     arbitrary/optional/path/to/workerName
+            //
             // We look for a file:
             //
-            //                    <workerPath>/<workerName>.php
+            //     <workerPath>/<workerName>.php
             //
-            //                If it's there, we include it, and then create
-            //                an object and run it like:
+            // If it's there, we include it, and then create an object and run
+            // it like:
             //
-            //                    $worker = new $workerName( $job );
-            //                    $worker->safeRun( );
+            //     $worker = new $workerName( $job );
+            //     $worker->safeRun( );
             //
-            // The optional path info allows for jobs to be scheduled selectively. I.e., you may have separate jobs
-            // scheduled as production/jobName and staging/jobName, with a WorkerManager in each environment looking for
-            // each path.
+            // The optional path info allows for jobs to be scheduled
+            // selectively.  For example, you may have separate jobs scheduled
+            // as production/jobName and staging/jobName, with a WorkerManager
+            // in each environment looking for each path.
             $job = $response['body'];
             $parts = explode('/', $job['name']);
             $jobParts = explode('?', $job['name']);
@@ -262,6 +274,7 @@ try {
 } catch (Exception $e) {
     $message = $e->getMessage();
     $logger->alert('BedrockWorkerManager.php exited abnormally', ['exception' => $e]);
+    echo "Error: $message\r\n";
 }
 
 // We wait for all children to finish before dying.
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -144,9 +144,9 @@ try {
                 break 2;
             }
 
-            // Query the server for a job
+            // Ready to get a new job
             try {
-                // Call GetJob
+                // Query the server for a job
                 $response = $jobs->getJob($jobName, 60 * 1000); // Wait up to 60s
             } catch (Exception $e) {
                 // Try again in 60 seconds
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -16,7 +16,7 @@ use Expensify\Bedrock\Jobs;
  * After N cycle in the loop, we exit
  * If the versionWatchFile modified time changes, we stop processing new jobs and exit after finishing all running jobs.
  *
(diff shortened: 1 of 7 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -71,34 +71,37 @@
 Client::configure($bedrockConfig);
 
 $logger = Client::getLogger();
+$logger->info('Starting BedrockWorkerManager', ['maxLoopIteration' => $maxLoopIteration]);
```
quinthar: Just moving this line further up to keep the logging initialization code as early as possible.

bin/BedrockWorkerManager.php
```diff
@@ -71,34 +71,37 @@
...
+    // Loop a finite number of times and then self destruct.  This is to guard
+    // against memory leaks, as we assume there is some other script that will
+    // restart this when it dies.
     $loopIteration = 0;
     while (true) {
+        // Is it time to self destruct?
         if ($loopIteration === $maxLoopIteration) {
             $logger->info("We did all our loops iteration, shutting down");
-            exit(0);
+            break;
```
quinthar: `exit(0)` skips the "wait for children to finish" code at the end.

bin/BedrockWorkerManager.php
```diff
@@ -71,34 +71,37 @@
...
-
         $loopIteration++;
         $logger->info("Loop iteration", ['iteration' => $loopIteration]);
 
         // Step One wait for resources to free up
         while (true) {
             // Get the latest load
+            if (!file_exists('/proc/loadavg')) {
+                throw new Exception('are you in a chroot?  If so, please make sure /proc is mounted correctly');
+            }
```
quinthar: Moving the `/proc/loadavg` check here to keep it next to the `sys_getloadavg()` call that depends on it.

bin/BedrockWorkerManager.php
```diff
@@ -109,21 +112,28 @@
...
-            // php's filemtime results are cached, so we need to clear that cache or we'll be getting a stale modified time.
+            // Watch a version file that will cause us to automatically shut
+            // down if it changes.  This enables triggering a restart if new
+            // PHP is deployed.
+            //
+            // Note: php's filemtime results are cached, so we need to clear
+            //       that cache or we'll be getting a stale modified time.
             clearstatcache(true, $versionWatchFile);
-            $newVersionWatchFileTimestamp = $versionWatchFile && file_exists($versionWatchFile) ? filemtime($versionWatchFile) : false;
+            $newVersionWatchFileTimestamp = ($versionWatchFile && file_exists($versionWatchFile)) ? filemtime($versionWatchFile) : false;
```
quinthar: Just clarifying what this means with parenthesis as the precedence of the operators wasn't obvious to me at first.

bin/BedrockWorkerManager.php
```diff
@@ -245,10 +259,13 @@
...
 } catch (Exception $e) {
     $message = $e->getMessage();
     $logger->alert('BedrockWorkerManager.php exited abnormally', ['exception' => $e]);
 }
+
+// We wait for all children to finish before dying.
+$logger->info('Stopping BedrockWorkerManager, waiting for children');
+$status = null;
+pcntl_wait($status);
+$logger->info('Stopped BedrockWorkerManager');
```
quinthar: Moving all this logic to the end of the outer loop, so it runs regardless of whether we're doing a graceful or abnormal shutdown.

bin/BedrockWorkerManager.php
```diff
@@ -71,34 +71,37 @@
 Client::configure($bedrockConfig);
```
quinthar: @iwiznia - Why do you manually copy all the values over to `$bedrockConfig` before passing it in, rather than just passing in `$options`?

bin/BedrockWorkerManager.php
```diff
@@ -16,21 +16,32 @@
...
-if (!$maxLoopIteration) {
-    $maxLoopIteration = 1000;
+$workerPath = @$options['workerPath'];
+if (!$workerPath) {
+    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency>]\r\n";
+    exit(1);
 }
+
+// Add defaults
+$jobName = @$options['jobName'] ?: '*'; // Process all jobs by default
```
quinthar: Moved this up closer to the usage description and updated the defaults to work out of the box with a minimum number of command line parameters.

bin/BedrockWorkerManager.php
```diff
@@ -59,46 +70,51 @@
...
-    // Begin the infinite loop
-    $loopIteration = 0;
+    // If --maxIterations is set, loop a finite number of times and then self
+    // destruct.  This is to guard against memory leaks, as we assume there is
+    // some other script that will restart this when it dies.
+    $iteration = 0;
     while (true) {
-        if ($loopIteration === $maxLoopIteration) {
+        // Is it time to self destruct?
+        if ($maxIterations > 0 && $iteration >= $maxIterations) {
```
quinthar: Renamed these variables to just use `iterations` rather than `loopIterations`.

bin/BedrockWorkerManager.php
quinthar: @iwiznia Are you sure this shutdown code does anything?

sample/SampleWorker.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Sample BedrockWorkerManager worker class.
+ */
+class SampleWorker extends BedrockWorker
```
quinthar: Adding this to provide a quick test and demonstration harness for how BWM works.  In doing so I discovered that the `Bedrock-PHP` repo is incomplete, as `BedrockWorker` isn't included here.  We need to extract that from the Expensify codebase and include in `Bedrock-PHP/src` instead.

sample/SampleWorker.php
```diff
@@ -0,0 +1,15 @@
...
+{
+    /**
+     * This is called once to run the worker. The thread will be forked before
+     * this is called, and after this function exits the process will exit.
+     */
+    public function run()
+    {
+        Log::info("Running SampleWorker for '{$this->job['name']}'");
```
quinthar: This is a call to the Expensify `Log` class -- need to remove.

bin/BedrockWorkerManager.php
```diff
@@ -16,89 +16,77 @@
...
 }
 
-Client::configure($bedrockConfig);
+// Add defaults
+$jobName = @$options['jobName'] ?: '*'; // Process all jobs by default
+$maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
+$maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
+
+// Configure the Bedrock client with these command-line options
+Client::configure($options);
```
quinthar: I didn't see the value of `$bedrockConfig` as it seems to just be a direct copy of `$options` so I removed it.  Please let me know if there was some hidden intent behind it, otherwise I'm assuming it was cruft over (eg, maybe it previously had some kind of value, but some change obviated that value and this was just left over as an obsolete relic of the past).

src/Client.php
```diff
@@ -73,6 +74,15 @@ public static function getStats()
     }
 
     /**
+     * Sets the global requestID, which is used if no instance requestID is given.
+     */
+    public static function setGlobalRequestID($globalRequestID)
+    {
+        // Override the global requestID
+        self::$globalRequestID = $globalRequestID;
```
quinthar: Note: this is called in `Web-Expensify/_config.php` and `Web-Expensify/lib/Logger.php`

src/Jobs.php
```diff
@@ -185,9 +186,10 @@ public function updateJob($jobID, $data, $repeat = null)
         return $this->call(
             "UpdateJob",
             [
-                "jobID"    => $jobID,
-                "data"     => $data,
-                "repeat"   => $repeat,
+                "jobID"      => $jobID,
+                "data"       => $data,
+                "repeat"     => $repeat,
+                "idempotent" => "true",
```
quinthar: This header just indicates that it's safe to execute this command twice, as the result will be the same.

bin/BedrockWorkerManager.php
```diff
@@ -16,89 +16,77 @@
...
+if (!$workerPath) {
+    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency>]\r\n";
+    exit(1);
 }
 
-Client::configure($bedrockConfig);
+// Add defaults
+$jobName = @$options['jobName'] ?: '*'; // Process all jobs by default
+$maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
+$maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
```
quinthar: I had a brief conversation with @cead22 about the "error control operator".  I understand that we discourage it (though I don't understand why).  Regardless, so far as I can tell we don't explicitly forbid it -- just discourage it -- which I'm interpreting as meaning it's up to the programmer's judgement.  And in my judgement, the code above is far more readable than the alternative:

```
$jobName = isset($options['jobName']) ? $options['jobName'] : '*';
$maxLoad = (isset($options['maxLoad'] && floatval($options['maxLoad'])) ? floatval($options['maxLoad']) : 1.0;
$maxIterations = (isset($options['maxIterations'] && intval($options['maxIterations'])) ? intval($options['maxIterations']) ? -1;
```

bin/BedrockWorkerManager.php
iwiznia: I don't think we need to do this check every time the loop runs.

bin/BedrockWorkerManager.php
iwiznia: Yes.

sample/SampleWorker.php
iwiznia: If you are not moving `BedrockWorker` to this repo nor removing the `Log` call, I say we just leave the SampleWorker in the web repo for now and move it whenever we do that.

sample/test.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
iwiznia: I think we should write real tests (with PHPUnit) and also set up travis to run them (to do this we also need to download the latest bedrock build from the public repo).

sample/test.sh
```diff
@@ -0,0 +1,70 @@
...
+' | nc localhost 8888 | head -n 1`
+if [[ "$RESULT" != 200* ]]
+then
+    echo "ERROR: Cleanup failed ($RESULT)"
+    exit
+fi
+
+
+# -----------------------
+echo 'Creating 2 jobs...'
```
iwiznia: 1 job

src/Client.php
```diff
@@ -170,11 +182,31 @@ public function call($method, $headers = [], $body = '')
...
 
-        // Do the request.  This is split up into separate functions so we can
-        // profile them independently -- useful when diagnosing various network
-        // conditions.
-        $this->sendRawRequest($rawRequest);
-        $response = $this->receiveRawResponse();
+        // Try idempotent requests up to three times, everything else only once
+        $numTries = (@$headers['idempotent'] === 'true') ? 3 : 1;
+        $response = null;
+        while($numTries-- && !$response)
```
iwiznia: Style: all these `{` should be on the same line.

src/Client.php
```diff
@@ -170,11 +182,31 @@ public function call($method, $headers = [], $body = '')
...
+                $this->sendRawRequest($rawRequest);
+                $response = $this->receiveRawResponse();
+            }
+            catch(ConnectionFailure $e)
+            {
+                // Failed to connect.  Are we retrying?
+                if ($numTries) {
+                    $this->getLogger()->warning("Failed to connect, send, or receive request; retrying $numTries more times", ['message' => $e->getMessage()]);
+                } else {
+                    $this->getLogger()->warning("Failed to connect, send, or receive request; not retrying", ['message' => $e->getMessage()]);
```
iwiznia: Log `error` instead of warning?

src/Client.php
```diff
@@ -146,6 +156,8 @@ public function call($method, $headers = [], $body = '')
         // Include the requestID for logging purposes
         if ($this->requestID) {
             $headers['requestID'] = $this->requestID;
+        } else if (self::$globalRequestID) {
```
iwiznia: Does this work? I think `$this->requestID` will always be set, won't it?

src/Client.php
```diff
@@ -170,11 +182,31 @@ public function call($method, $headers = [], $body = '')
...
+            // Catch any connection failures and retry, but ignore non-connection failures.
+            try
+            {
+                // Do the request.  This is split up into separate functions so we can
+                // profile them independently -- useful when diagnosing various network
+                // conditions.
+                $this->sendRawRequest($rawRequest);
+                $response = $this->receiveRawResponse();
+            }
+            catch(ConnectionFailure $e)
```
iwiznia: Not sure I get this retry logic and the commands being marked as `idempotent`. 
This catch, will catch ConnectionFailure errors thrown either in `reconnect` or in `recv`. The ones in reconnect I'm not worried about, but the ones in `recv` I think require additional logic. So, when it fails on `recv` there's a chance that bedrock did actually process the command, but we failed to get the response from that command back for whatever reason. If this is the case, then:
- We can't retry a `GetJob` command (well, we can, but we'll get a different job back and the one that we got previously would be stuck in RUNNING state).
- We can retry a finish/retry/fail command, but we need to handle the response differently in case of a retry, because when retried those commands will return an error 405 (since the job needs to be in RUNNING state).

src/Client.php
```diff
@@ -170,11 +182,31 @@ public function call($method, $headers = [], $body = '')
         $rawRequest .= "\r\n";
         $rawRequest .= $body;
 
-        // Do the request.  This is split up into separate functions so we can
-        // profile them independently -- useful when diagnosing various network
-        // conditions.
-        $this->sendRawRequest($rawRequest);
-        $response = $this->receiveRawResponse();
+        // Try idempotent requests up to three times, everything else only once
+        $numTries = (@$headers['idempotent'] === 'true') ? 3 : 1;
```
iwiznia: Why use `"true"` instead of `true`?

bin/BedrockWorkerManager.php
quinthar: Oh, sweet!

bin/BedrockWorkerManager.php
quinthar: Agreed we don't need to, but I think it's non-harmful to do it (eg, it's not a super expensive call given the relatively low the frequency of this loop), and makes the code read a lot cleaner to keep it where it's used.

sample/SampleWorker.php
quinthar: How about I just move SampleWorker to Bedrock-PHP as part of this PR, and remove its call to `Log::`, but we leave `BedrockWorker` where it is for now.  So I'll make partial progress.

quinthar: Done.

sample/test.sh
quinthar: I agree that would be cool, but that's a larger effort than I wanted to take right now.

sample/test.sh
quinthar: Actually it creates 2 jobs, back to back.

src/Client.php
quinthar: No idea why I did that.  Fixing.

src/Client.php
quinthar: Yes, but I think that's already accounted for by flagging which commands are idempotent:

* `GetJobs` is *kind* of idempotent.  Yes, if it fails, it'll leave a job `RUNNING` forever.  But that's not BWM's problem -- all it's going to do is call `GetJobs` again.  So retrying `GetJobs` is safe.
* Something like `QueryJob` is fine to retry.
* Even something like `FinishJob` or `RetryJob` are fine.  Either they work right, or they return a "this has already been finished" error.  Either way, there's no harm in retrying.
* However, something like `CreateJob` *cannot* be safely retried, as you might be trying to create a job that executes exactly once, so retrying could cause it to happen twice (or more).

Ultimately, flag it as `idempotent: true` if it's safe to retry -- and in almost all cases, it is.

src/Client.php
quinthar: Hm, well `warning` implies "this is something bad but not fatal", `error` implies "this is fatal and non-recoverable" to me.

quinthar:
Ready for re-review.

src/Client.php
iwiznia: > Even something like FinishJob or RetryJob are fine. Either they work right, or they return a "this has already been finished" error. Either way, there's no harm in retrying.

This is the part that I think is wrong. So, you are going to retry and it's going to respond with `a "this has already been finished" error`, when that happens, we need to treat that error as a success instead or BWM will try to mark the job as failed (and fail in doing so too)

src/Client.php
iwiznia: That's my point, when you retry, it's a warning, when you stopped retrying, is an error.

sample/test.sh
iwiznia: Alright.
FYI: moving the whole code to a class, like my refactoring PR did, would allow us to write proper integration tests that can be included in the repo and run by travis.

src/Client.php
quinthar: Ahh got it, will do.

src/Client.php
quinthar: How `REQUEST_ID` works today:
----
Actually no, it's not set by default.  It's [set to `null` in `configure()`](https://github.com/Expensify/Bedrock-PHP/blob/master/src/Client.php#L34) by default, and that default is [not overridden in `BedrockWorkerManager.php`](https://github.com/Expensify/Bedrock-PHP/blob/master/bin/BedrockWorkerManager.php#L71).  Indeed, I don't actually think that `$this->requestID` is set *anywhere*.

The closest we seem to do is this [totally insane hack](https://github.com/Expensify/Web-Expensify/blob/master/lib/Logger.php#L69), but that only sets the `requestID` for `Log::info()`: it doesn't tell Bedrock to send it.

And the only reason Auth gets it is because we don't use `Bedrock\Client` *and* don't look at `Log`.  Rather, we just use [set it equal to the global `REQUEST_ID` constant directly](https://github.com/Expensify/Web-Expensify/blob/master/lib/Auth.php#L621).

What this PR does:
----
So, I've updated [`_config.php` to set `requestID`](https://github.com/Expensify/Web-Expensify/pull/17025/files#diff-8a7315557cd9f672e36f7e8f0ce2c29eR33) so it's automatically set at the very start of the script, and also setting it [inside the crazy hack](https://github.com/Expensify/Web-Expensify/pull/17025/files#diff-aac2b4bbc948c6d3cbd35384ee02e473R72).

Unfortunately, I'm realizing now what sucks about this is we will *not* update the `REQUEST_ID` constant -- it'll still be set to the `REQUEST_ID` of the parent BWM process.  So this PR will create traceability through Bedrock, but not into Auth.

Proposing a better solution:
----
So I think the *real* solution would be to:

1. Replace the `REQUEST_ID` constant with a `$REQUEST_ID` global variable, and then just use that everywhere.  (I'm not sure why we're setting it separately inside `Log` -- we should just use the global.)  This would be pretty straightforward -- it's not actually used in that many places:

```
Davids-MacBook-Pro:Web-Expensify dbarrett$ grep -s -r REQUEST_ID *
_config.php: * REQUEST_ID yet. This will be set with the logID request parameter if it is
_config.php:if (!defined('REQUEST_ID') &&
_config.php:    define('REQUEST_ID', $_REQUEST['logID']);
_config.php:    define('REQUEST_ID', Log::generateID());
_config.php:Log::init(REQUEST_ID);
_config.php:Client::setGlobalRequestID(REQUEST_ID);
_config.php:    newrelic_add_custom_parameter('REQUEST_ID', REQUEST_ID);
_libfop.php:    $content['logID'] = REQUEST_ID;
_libfop.php:    $params['requestID'] = REQUEST_ID;
_tasks/Hit.class.php:            "logID"      => REQUEST_ID,
_testify/receiptUpload/testAll.php:    $parameters['logID'] = REQUEST_ID;
_testify/receiptUpload/testAll.php:    $parameters['email'] = 'AMEX_'.REQUEST_ID;
_testify/smartscan/autoScan.php:        "logID"      => REQUEST_ID,
api/amex/v1/lib/net.class.php:        $parameters['logID'] = REQUEST_ID;
api/amex/v1/lib/net.class.php:        $parameters['email'] = 'AMEX_'.REQUEST_ID;
externalLib/amazon-sdk-2.3.2/Aws/AutoScaling/Enum/ScalingActivityStatusCode.php:    const WAITING_FOR_SPOT_INSTANCE_REQUEST_ID = 'WaitingForSpotInstanceRequestId';
lib/Auth.php:        $headers['requestID'] = REQUEST_ID;
lib/Event.php:        $safeRequestID = DB_Analytics::escapeWithoutConnection(REQUEST_ID);
lib/Notification/Standard.php:        $params['requestID'] = REQUEST_ID;
lib/Report/Submit/Utils.php:        TechnicalContactNotifier::addMessage('Request ID '.REQUEST_ID);
Davids-MacBook-Pro:Web-Expensify dbarrett$ 
````

2. Alternatively, we use something like [runkit_constant_redefine()](http://www.php.net/manual/en/function.runkit-constant-redefine.php) to just redefine it when we `fork()`.  This requires installing a custom package, however.

3. A third option would be to remove `REQUEST_ID` entirely... but I'm not sure really how to do that.

I think I lean toward (1) above.  Thoughts?

Cc: @cead22

Comment (cead22) [inline src/Client.php]: I agree that 1. seems like the best option. 2. has the drawbacks of requiring a package and maybe a lesser one that we're redefining a constant which can be confusing. As for 3. I don't even know what that would entail

new messages

File: src/Client.php [hunk 22]
```diff
@@ -84,12 +84,11 @@ public static function getStats()
      * @param int             $readTimeout       Timeout to use when reading
      * @param LoggerInterface $logger            Class to use for logging
      * @param StatsInterface  $stats             Class to use for statistics tracking
-     * @param string          $requestID         RequestID to send to bedrock for consolidated logging
      * @param string          $writeConsistency  The bedrock write consistency we want to use
      *
      * @throws BedrockError
      */
-    public function __construct($host = null, $port = null, $failoverHost = null, $failoverPort = null, $connectionTimeout = null, $readTimeout = null, LoggerInterface $logger = null, StatsInterface $stats = null, $requestID = null, $writeConsistency = null)
+    public function __construct($host = null, $port = null, $failoverHost = null, $failoverPort = null, $connectionTimeout = null, $readTimeout = null, LoggerInterface $logger = null, StatsInterface $stats = null, $writeConsistency = null)
```
quinthar: FYI, so far as I can tell nobody ever uses this constructor for anything but the defaults.  My sense is this many attributes in the constructor is pretty excessive, especially since we have a static `configure()` function.  I'd vote we just remove the constructor, or remove `configure()`.  I'm not sure we're getting much value by having both.

PS: Why is this named `Client` rather than just the more obvious `Bedrock`?  I understand the full class is `Expensify\Bedrock\Client`, and I even understand that other libraries (eg, Intercom) use the `Client` moniker.  But it makes searching for it really hard because the fully-qualified name is never used:

```
Davids-MacBook-Pro:Web-Expensify dbarrett$ grep -i -r -s 'new client' * | grep -i bedrock | grep .php: | grep -v vendor/ | grep -v externalLib/
_devportal/tools/bedrockJobManager/api.php:$bedrock = new Client();
_devportal/tools/harvester/personalAutoHarvest.php:        $bedrock = new Client();
_tasks/ParseResult.class.php:        $bedrock = new Client();
_tasks/ScanningAPI.class.php:        $bedrock = new Client();
_tests/integration/api/Policy/Connection/SetAutoSyncTest.php:        $bedrock = new Client();
_tests/integration/api/Policy/Connection/SetAutoSyncTest.php:        $bedrock = new Client();
_tests/integration/api/Policy/Connection/SetAutoSyncTest.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/ChatBotTestUtils.php:        $bedrock = new Client();
_tests/integration/IntegrationTestUtils.php:        $db = new \Expensify\Bedrock\DB(new Client());
_tests/integration/IntegrationTestUtils.php:        $db = new \Expensify\Bedrock\DB(new Client());
_tests/integration/IntegrationTestUtils.php:        $db = new \Expensify\Bedrock\DB(new Client());
_tests/integration/lib/Bedrock/JobsTest.php:        self::$jobs = new Jobs(new Client());
_tests/integration/lib/Bedrock/JobsTest.php:        $db = new DB(new Client());
_tests/integration/lib/Bedrock/JobsTest.php:        $db = new DB(new Client());
_tests/integration/lib/Bedrock/JobsTest.php:        $db = new DB(new Client());
_tests/integration/lib/Bedrock/ParentJobsTest.php:        $jobs = new Jobs(new Client());
_tests/integration/lib/Bedrock/ParentJobsTest.php:        $jobs = new Jobs(new Client());
_tests/integration/lib/Bedrock/ParentJobsTest.php:        $jobs = new Jobs(new Client());
_tests/integration/lib/Bedrock/ParentJobsTest.php:        $jobs = new Jobs(new Client());
_tests/integration/lib/Bedrock/ParentJobsTest.php:        $jobs = new Jobs(new Client());
_tests/integration/lib/Bedrock/PriorityTest.php:        $jobs = new Jobs(new Client());
_tests/integration/lib/ChatBot/APITest.php:        $bedrock = new Client();
_tests/integration/lib/Tab/UtilsTest.php:        self::$bedrock = new DB(new Client());
lib/ChatBot/ChatBot.php:        $bedrock = new Client();
lib/ChatBot/Store/Chat.php:        $bedrock = new Client();
lib/ChatBot/Store/Chat.php:        $bedrock = new Client();
lib/ChatBot/Store/Event.php:            $bedrock = new Client();
lib/ChatBot/Store/Event.php:        $bedrock = new Client();
lib/ChatBot/Store/Event.php:        $bedrock = new Client();
lib/ChatBot/Store/Input.php:        $bedrock = new Client();
lib/ChatBot/Store/Input.php:        $bedrock = new Client();
lib/ChatBot/Store/Input.php:            $bedrock = new Client();
lib/ChatBot/Store/Input.php:        $bedrock = new Client();
lib/ChatBot/Store/State.php:        $bedrock = new Client();
lib/DomainSummary.php:            $bedrock = new Client();
lib/Notification/Standard.php:                $bedrock = new Client();
lib/Policy.php:            $bedrock = new Client();
lib/Policy.php:                    $bedrock = new Client();
lib/PolicyAPI.php:        $bedrock = new Client();
lib/Report/Reward.php:            $bedrockJobs = new Jobs(new Client());
lib/ReportUtils.php:        $bedrock = new Client();
script/bwm/EmailSendLogger.php:        $bedrock = new Client();
script/bwm/PolicyHarvester.php:        $bedrock = new Client();
script/bwm/PolicyHarvester.php:        $bedrock = new Client();
script/bwm/PolicySummaryEmail.php:            $bedrock = new Client();
script/bwm/UserHarvester.php:        $bedrock = new Client();
script/createDomainWorkers.php:$bedrock = new Client();
Davids-MacBook-Pro:Web-Expensify dbarrett$ 
```

sample/test.sh
quinthar: oh, and to clarify, my intention here is not to create a full test script.  Rather, I want the Bedrock-PHP repo to have a trivial example that future developers can study to understand how the system works.  So the point isn't to exercise all code paths, but merely to demonstrate how to use BWM in the real world.

quinthar: I'm going to rename this to be `demo.sh` accordingly, to set the right expectations.

quinthar:
@iwiznia Ready for re-review

bin/BedrockWorkerManager.php
```diff
@@ -16,89 +16,77 @@
...
-$versionWatchFile = isset($options['versionWatchFile']) ? $options['versionWatchFile'] : null;
-$workerPath = isset($options['workerPath']) ? $options['workerPath'] : null;
-if (!$jobName || !$maxLoad || !$workerPath) {
-    throw new Exception('Usage: sudo -u user php ./bin/BedrockWorkerManager.php --jobName=<jobName> --workerPath=<workerPath> --maxLoad=<maxLoad> [--host=<host> --port=<port> --maxIterations=<loopIteration> --writeConsistency=<consistency>]');
-}
-if ($maxLoad <= 0) {
-    throw new Exception('Maximum load must be greater than zero');
+// Parse the command line and verify the required settings are provided
+$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::']);
+$workerPath = @$options['workerPath'];
```
iwiznia: `$options['workerPath'] ?? null`

bin/BedrockWorkerManager.php
```diff
@@ -16,89 +16,77 @@
...
+if (!$workerPath) {
+    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency>]\r\n";
+    exit(1);
 }
 
-Client::configure($bedrockConfig);
+// Add defaults
+$jobName = $options['jobName'] ?? '*'; // Process all jobs by default
+$maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
```
iwiznia: `floatval($options['maxLoad'] ?? 1.0)`

bin/BedrockWorkerManager.php
```diff
@@ -16,89 +16,77 @@
...
+$maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
```
iwiznia: `intval($options['maxIterations'] ?? -1)`

bin/BedrockWorkerManager.php
```diff
@@ -16,89 +16,77 @@
...
+// Configure the Bedrock client with these command-line options
+Client::configure($options);
+
+// Prepare to use the host logger, if configured
 $logger = Client::getLogger();
-$stats = Client::getStats();
+$logger->info('Starting BedrockWorkerManager', ['maxIterations' => $maxIterations]);
+
+// If --versionWatch is enabled, begin watching a version file for changes
+$versionWatchFile = @$options['versionWatchFile'];
```
iwiznia: `$options['versionWatchFile'] ?? null`

src/Client.php
```diff
@@ -170,11 +168,28 @@ public function call($method, $headers = [], $body = '')
         $rawRequest .= "\r\n";
         $rawRequest .= $body;
 
-        // Do the request.  This is split up into separate functions so we can
-        // profile them independently -- useful when diagnosing various network
-        // conditions.
-        $this->sendRawRequest($rawRequest);
-        $response = $this->receiveRawResponse();
+        // Try idempotent requests up to three times, everything else only once
+        $numTries = @$headers['idempotent'] ? 3 : 1;
```
iwiznia: `isset($headers['idempotent']) ? 3 : 1`

src/Client.php
```diff
@@ -170,11 +168,28 @@ public function call($method, $headers = [], $body = '')
...
+        while($numTries-- && !$response) {
+            // Catch any connection failures and retry, but ignore non-connection failures.
+            try {
+                // Do the request.  This is split up into separate functions so we can
+                // profile them independently -- useful when diagnosing various network
+                // conditions.
+                $this->sendRawRequest($rawRequest);
+                $response = $this->receiveRawResponse();
+            }
+            catch(ConnectionFailure $e) {
```
iwiznia: NAB: this should be on the same line as the `}` above.

src/Client.php
```diff
@@ -170,11 +168,28 @@ public function call($method, $headers = [], $body = '')
...
 
-        // Do the request.  This is split up into separate functions so we can
-        // profile them independently -- useful when diagnosing various network
-        // conditions.
-        $this->sendRawRequest($rawRequest);
-        $response = $this->receiveRawResponse();
+        // Try idempotent requests up to three times, everything else only once
+        $numTries = @$headers['idempotent'] ? 3 : 1;
+        $response = null;
+        while($numTries-- && !$response) {
```
iwiznia: [This comment](https://github.com/Expensify/Bedrock-PHP/pull/15#discussion_r111559831) wasn't answered.

bin/BedrockWorkerManager.php
```diff
@@ -166,24 +160,40 @@
...
                         'extraParams' => $extraParams,
                     ]);
-
+                    if (isset($GLOBALS['REQUEST_ID'])) {
+                        // Reset the REQUEST_ID and re-log the line so we see
+                        // it when searching for either the parent and child
+                        // REQUEST_IDs.
+                        mt_srand(getmypid());
+                        $GLOBALS['REQUEST_ID'] = substr(base64_encode(sha1(mt_rand())),0,6); // random 6 character ID
+                        $logger->info("Fork succeeded, child process, running job", [
```
iwiznia: Can you change this text a bit so it's easier to search for parent vs child?

bin/BedrockWorkerManager.php
quinthar: Per offline discussion, this is NAB

bin/BedrockWorkerManager.php
quinthar: Per offline discussion, this is NAB

quinthar: Actually... on this does the `??` just implicitly do the error suppression on the left expression?  So it'll work even if inside a function?  That's pretty slick.

bin/BedrockWorkerManager.php
quinthar: Per offline discussion, this is NAB

bin/BedrockWorkerManager.php
iwiznia: `$options['maxLoad'] ?? 1.0` is equivalent to `(isset($options['maxLoad']) ? isset($options['maxLoad'] : 1.0)`

PR #16, opened 2017-04-22
More randomness

quinthar:
Fixes https://github.com/Expensify/Expensify/issues/50915


mea36:
You need to update the version number [here](https://github.com/Expensify/Bedrock-PHP/blob/master/composer.json#L5)

PR #20, opened 2017-05-02
Use idempotent parameter for CreateJob

francoisl:
@iwiznia please review

When calling `CreateJob`, set `idempotent` based on `$unique`. This will allow any request that failed due a `ConnectionFailure` to be automatically retried if `$unique = true`.
Given that when creating jobs with `$unique = true`, Bedrock returns data about any existing job with the same name if it exists, there's no risk to accidentally create another job with the same name.


quinthar:
Clever!

PR #22, opened 2017-06-13
Append a semicolon to queries if none exists

mea36:
cc @cead22 
Let's append `;` to a query if it does not exist

This is needed for https://github.com/Expensify/Bedrock/pull/215

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -39,6 +39,7 @@ class DB extends Plugin
      */
     public function query($sql)
     {
+        $sql = substr($sql, -1) === ";" ? $sql : $sql.";";
         $response = new Response($this->client->call(
             "Query",
             [
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "1.0.16",
+    "version": "1.0.17",
     "authors": [
         {
             "name": "Expensify",


src/DB.php
```diff
@@ -39,6 +39,7 @@ class DB extends Plugin
      */
     public function query($sql)
     {
+        $sql = substr($sql, -1) === ";" ? $sql : $sql.";";
```
flodnv: My 2 cents is that this shouldn't be the responsibility of the DB layer. It should be the responsibility of callers to know how to write valid queries. That being said, I am not adding myself as a reviewer.

mea36: @cead22 and I went back and forth on this one for a bit.  We ended up making the change here as opposed to `Web-Expensify` because this is an interface with `Bedrock` and therefore should simplify communication with `Bedrock`.
The argument that won me over was 
>you don't send connectionTimeout with every request when you use the php client

iwiznia: Huh? That's a bad example, we do send `connectionTimeout` on every request. It's part of the configuration and it's sent every time. I think you are comparing apples to oranges (hate this expression).

mea36: Haha, yes, it's sent on every request but it doesn't need to explicitly be set by the user, and that was the point.  
Yes, a semicolon is something that a query needs, but if the user doesn't add, we can

cead22: To clarify, this is not the DB layer, this is still the client.

A better analogy or example that supports this decision is: sequel pro doesn't require you to have a `;` at the end of a query.

PR #23, opened 2017-06-14
Temporary BWM load handler

mattabullock:
**Problem**
BWM-PHP kicks off multiple forks of itself to connect to bedrock to query for work (the query itself is blocking - when BWM has a job ready, it’ll then respond to the request).
 
We have a substantial amount of jobs that are queued to be worked on, and they cannot all kick off at once or else we will effectively DDOS ourselves.

**Solution**
Use a command-line argument that caps the number of forks that can exist at any moment to a configurable number.

**Tests**
1. Create 10 jobs with a 5 second sleep.
2. Run bwm with `--maxNumberWorkerThreads=1`
3. Tail logs, confirm number of worker threads is never above 1.
4. Repeat steps 1-3 with `--maxNumberWorkerThreads=5`.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -29,14 +29,14 @@ if (php_sapi_name() !== "cli") {
 }
 
 // Parse the command line and verify the required settings are provided
-$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::', ‘maxNumberWorkerThreads::’]);
+$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::', 'maxNumberWorkerThreads::']);
 
 // Store parent ID to determine if we should continue forking
 $parentID = getmypid();
 
 $workerPath = @$options['workerPath'];
 if (!$workerPath) {
-    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency>]\r\n";
+    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency> --maxNumberWorkerThreads=<maxNumberWorkerThreads>]\r\n";
     exit(1);
 }
 
@@ -44,7 +44,7 @@ if (!$workerPath) {
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
 $maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
 $maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
-$maxNumberWorkerThreads = intval($options[‘maxNumberWorkerThreads’]) ?? 5; // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
+$maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads']) ?? 5; // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
 
 // Configure the Bedrock client with these command-line options
 Client::configure($options);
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -29,7 +29,11 @@
 }
 
 // Parse the command line and verify the required settings are provided
-$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::']);
+$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::', ‘maxNumberWorkerThreads::’]);
```
mcnamamj: wrong style tick

bin/BedrockWorkerManager.php
```diff
@@ -40,6 +44,7 @@
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
 $maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
 $maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
+$maxNumberWorkerThreads = intval($options[‘maxNumberWorkerThreads’]) ?? 5; // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
```
mcnamamj: wrong style tick/quote

capezzbr: use `'` instead of ``` ?

bin/BedrockWorkerManager.php
```diff
@@ -81,18 +86,24 @@
...
                 throw new Exception('are you in a chroot?  If so, please make sure /proc is mounted correctly');
             }
+
+            // Check if we can fork based on our hard-coded limit
+            exec("pgrep -P $parentID", $output);
+
+            // Check if we can fork based on the load of our webservers
             $load = sys_getloadavg()[0];
-            if ($load < $maxLoad) {
+            if ($load < $maxLoad || count($output) < $maxNumberWorkerThreads) {
```
cead22: Log `count(output)` and `maxNumberWT` on both branches

bin/BedrockWorkerManager.php
```diff
@@ -81,18 +86,24 @@
...
                 $logger->info('Load is under max, checking for more work.', ['load' => $load, 'MAX_LOAD' => $maxLoad]);
```
capezzbr: not sure by the execution flow here, but perhaps reset `$forkIterations` here?

bin/BedrockWorkerManager.php
mattabullock: Yeah, I copied it from google docs and that's what it copied as.

bin/BedrockWorkerManager.php
zlailari: I think it's simpler where it is so we don't have to reset it and declare it in separate places

zlailari:
Updated

bin/BedrockWorkerManager.php
```diff
@@ -81,18 +86,24 @@
...
+        $forkIterations = 0;
         // Step One wait for resources to free up
         while (true) {
             // Get the latest load
             if (!file_exists('/proc/loadavg')) {
                 throw new Exception('are you in a chroot?  If so, please make sure /proc is mounted correctly');
             }
+
+            // Check if we can fork based on our hard-coded limit
+            exec("pgrep -P $parentID", $output);
```
mcnamamj: for future reference, this is a terrible variable name <3

bin/BedrockWorkerManager.php
```diff
@@ -81,18 +86,24 @@
         $iteration++;
         $logger->info("Loop iteration", ['iteration' => $iteration]);
 
+        $forkIterations = 0;
```
quinthar: Can you put this *under* the comment?  The pattern is:

```
// This is a comment summarizing what's below in English.
thisIsCode();

// After a blank line, another comment.  If you think it's too little
// to warrant summarizing, don't make a blank line -- consolidate
// code into meaningfully sized blocks, each with a meaningful
// comment.
```

bin/BedrockWorkerManager.php
```diff
@@ -29,17 +29,22 @@
 }
 
 // Parse the command line and verify the required settings are provided
-$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::']);
+$options = getopt('', ['host::', 'port::', 'failoverHost::', 'failoverPort::', 'maxLoad::', 'maxIterations::', 'jobName::', 'logger::', 'stats::', 'workerPath::', 'versionWatchFile::', 'writeConsistency::', 'maxNumberWorkerThreads::']);
+
+// Store parent ID to determine if we should continue forking
+$parentID = getmypid();
```
quinthar: This is an odd variable name -- in the scope of this function, it's not a "parent" ID: it's *our* ID.  Furthermore, I don't think this is an expensive call so I'd suggest just calling it where you need it rather than storing it in a variable.  The best named variable is the one that doesn't exist.

mattabullock:
Updated

bin/BedrockWorkerManager.php
```diff
@@ -29,17 +29,22 @@
...
-    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency>]\r\n";
+    echo "Usage: sudo -u user php ./bin/BedrockWorkerManager.php --workerPath=<workerPath> [--jobName=<jobName> --maxLoad=<maxLoad> --host=<host> --port=<port> --maxIterations=<iteration> --writeConsistency=<consistency> --maxNumberWorkerThreads=<maxNumberWorkerThreads>]\r\n";
     exit(1);
 }
 
 // Add defaults
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
 $maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
 $maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
+$maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads']) ?? 5; // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
```
fnwbr: image

fnwbr: This should be
```php
$maxNumberWorkerThreads = intval(@$options['maxNumberWorkerThreads']) ?: 5;
```

PR #24, opened 2017-06-16
Use intval after the null coalescing operator so the default can be properly set

zlailari:
The `intval` call was preventing the null coalescing operator from properly setting the default value

image

# Tests
1. Create 10 jobs with a 5 second sleep.
1. Run bwm with --maxNumberWorkerThreads=4
    - Tail logs, confirm number of worker threads is never above 4 (search for `Load is`) and there are no notices referencing `maxNumberWorkerThreads`
1. Repeat steps 1-2 without a default
    - Tail logs, confirm that the number of worker threads is never over the default (5) and there are no notices referencing `maxNumberWorkerThreads`

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -42,8 +42,8 @@ if (!$workerPath) {
 
 // Add defaults
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
-$maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
-$maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
+$maxLoad = floatval($options['maxLoad'] ?? 1.0); // Max load of 1.0 by default
+$maxIterations = intval($options['maxIterations'] ?? -1); // Unlimited iterations by default
 $maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads'] ?? 5); // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
 
 // Configure the Bedrock client with these command-line options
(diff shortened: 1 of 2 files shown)


chiragsalian:
ooh i was just doing the same thing

bin/BedrockWorkerManager.php
```diff
@@ -44,7 +44,7 @@
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
 $maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
 $maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
-$maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads']) ?? 5; // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
+$maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads'] ?? 5); // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
```
chiragsalian: btw i think this change needs to be `intval(@$options['maxNumberWorkerThreads']) ?? 5;`
adding the operator `@` would error control it.

zlailari: Using `??` won't output an `E_NOTICE` when you have an undefined variable, so we don't have to use the `@` operator. [Here's a stackoverflow post with more info](https://stackoverflow.com/questions/34571330/php-ternary-operator-vs-null-coalescing-operator)

chiragsalian: Ignore what i mentioned. As discussed 1:1 with ze'ev  `intval($options['maxNumberWorkerThreads'] ?? 5);` works well.

cead22:
Please add tests. There's a log line that logs this value so this should be trivial. Let me know if I can help

bin/BedrockWorkerManager.php
```diff
@@ -44,7 +44,7 @@
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
 $maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
 $maxIterations = intval(@$options['maxIterations']) ?: -1; // Unlimited iterations by default
```
fnwbr: NAB NAR: Can you adjust these two lines above to go in line with the one you modified and also use the `??` operator?

zlailari:
Updated and added tests

PR #28, opened 2017-07-17
[WIP] Add AIMD load handling algorithm to BWM

mattabullock:


quinthar:
Cool, looking good!

mattabullock:
Updated with full description and ready for review.

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,50 @@
...
+$maxSafeTime = intval($options['maxSafeTime']) ?: 0; // The maximum job time before we start paying attention
+$debugThrottle = $options['debugThrottle']; // Set to true to maintain a debug history
+$target = $minSafeJobs;
+$disableLoadHandler = $options['disableLoadHandler'];
+
+$localDB = new LocalDB($pathToDB);
+
+// Set up the database for the AIMD load handler.
+$query = "CREATE TABLE IF NOT EXISTS localJobs (
+    localJobID integer PRIMARY KEY NOT NULL,
```
iwiznia: Why not make this auto increment? I think it would simplify some things...

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
```
iwiznia: Declare strict types

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Expensify\Bedrock;
+
+use SQLite3;
+
```
iwiznia: Missing phpdoc

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+class LocalDB
+{
+    private $location;
```
iwiznia: Missing phpdoc

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Creates a LocalDB object and defines the file location.
+     *
+     * @param $location
```
iwiznia: Missing type annotation (here and in code too)

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Runs a read query on a local database.
+     *
+     * @param string $query
+     *
+     * @return array
+     */
+    public function read($query)
```
iwiznia: Missing type annotations

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $handle = new SQLite3($this->location);
```
iwiznia: I think you should do this in the constructor instead of on each read.

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+        unset($handle);
+        return $returnValue ?? null;
+    }
+
+    /**
+     * Runs a write query on a local database.
+     *
+     * @param string $query
+     */
+    public function write($query)
```
iwiznia: Missing type annotation

src/LocalDB.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $handle = new SQLite3($this->location);
```
iwiznia: Same here, do this in the constructor

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,50 @@
...
+$minSafeJobs = intval($options['minSafeJobs']) ?: 10;  // The minimum number of jobs before we start paying attention
+$maxSafeTime = intval($options['maxSafeTime']) ?: 0; // The maximum job time before we start paying attention
+$debugThrottle = $options['debugThrottle']; // Set to true to maintain a debug history
+$target = $minSafeJobs;
+$disableLoadHandler = $options['disableLoadHandler'];
+
+$localDB = new LocalDB($pathToDB);
+
+// Set up the database for the AIMD load handler.
+$query = "CREATE TABLE IF NOT EXISTS localJobs (
```
iwiznia: This should be inside a `if (!$disableLoadHandler)`

bin/BedrockWorkerManager.php
```diff
@@ -211,12 +237,17 @@
                     // that we automatically pick up new versions over the
                     // worker without needing to restart the parent.
                     include_once $workerFilename;
-                    $stats->benchmark('bedrockJob.finish.'.$job['name'], function () use ($workerName, $bedrock, $jobs, $job, $extraParams, $logger) {
+                    $stats->benchmark('bedrockJob.finish.'.$job['name'], function () use ($workerName, $bedrock, $jobs, $job, $extraParams, $logger, $localJobID, $localDB) {
                         $worker = new $workerName($bedrock, $job);
+                        $childPID = getmypid();
                         try {
+                            $localDB->write("INSERT INTO localJobs (localJobID, pid, jobID, jobName, started) values ($localJobID, $childPID, {$job['jobID']}, '{$job['name']}', DATETIME('now'));");
```
iwiznia: NAB: `VALUES`

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +309,55 @@
 $status = null;
 pcntl_wait($status);
 $logger->info('Stopped BedrockWorkerManager');
+
+// Determines whether or not we call GetJob and try to start a new job
```
iwiznia: Missing phpdoc

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +309,55 @@
...
+function safeToStartANewJob(LocalDB $localDB, int& $target, int $maxSafeTime, int $minSafeJobs, $debugThrottle, $logger)
```
iwiznia: Can we change this so that we don't pass the `$target` var by reference?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +309,55 @@
...
+{
+    // Have we hit our target job count?
+    $query = 'SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;';
+    $numActive = $localDB->read($query)[0];
+
+    if ($numActive < $target) {
+        // Still in a safe zone, don't worry about load
```
iwiznia: Can you add some logging on all these `return`s so it's easier for us to debug later?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +309,55 @@
...
+    if ($numFinished < $target*2) {
+        // Wait until we finish at least two batches of our target so we can evaluate its speed,
+        // before expanding the batch.
+        return false;
+    }
+
+    // Calculate the speed of the last 2 batches to see if we're speeding up or slowing down
+    $oldBatchTime = 0;
+    $oldBatchTimes = $localDB->read("SELECT cast(strftime('%s', ended)-strftime('%s', started) as int) from localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
+    if (is_array($oldBatchTimes)) {
```
iwiznia: 😕 isn't this always an array?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +309,55 @@
...
+        $oldBatchTime = array_sum($oldBatchTimes)/count($oldBatchTimes);
+    }
+
+    $newBatchTime = $localDB->read("SELECT sum(cast(strftime('%s', ended)-strftime('%s', started) as int))/count(*) from localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target;")[0];
```
iwiznia: NAB: `FROM`, `AS`

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +309,55 @@
...
+        // it's going roughly the same speed as the batch before.  This suggests that we
+        // haven't hit any serious bottleneck yet, so let's dial it up ever so slightly and see
+        // if speeds hold at the new target.
+        $target++;
+
+        // Also, blow away any data from more than two batches ago, because we don't
+        // look farther back than that and don't want to accumulate data infinitely.  However,
+        // this is very useful data to keep while debugging to analyze our throttle behavior.
+        if (!$debugThrottle) {
+            $localDB->write("DELETE FROM localjobs WHERE localjobid IN (SELECT localjobid FROM localjobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT -1 OFFSET $target*2);");
```
iwiznia: `LIMIT -1`??? What's this?

bin/BedrockWorkerManager.php
mattabullock: I tried but couldn't come up with a way for that to work since we have to update the job (using the localJobID) after the job runs and if you use an autoincrement there's no good way to do that since you've got multiple threads all writing at the same time.

bin/BedrockWorkerManager.php
mattabullock: Yeah that was a late hack I threw in. I could have it return multiple values and just unpack them in the function call if you think that's better.

bin/BedrockWorkerManager.php
mattabullock: You can't do offset without limit, and `LIMIT -1` is just unlimited. So this query deletes all finished localJobs older than `$target*2`.

bin/BedrockWorkerManager.php
flodnv: Yes, please always avoid using references, they can cause so many awful bugs

flodnv: also, please add types to `$debugThrottle, $logger` , thanks

bin/BedrockWorkerManager.php
iwiznia: hmmmmm, do all threads use the same instance of the sqlite object? If not, we can use the `lastInsertRowID` method safely.

bin/BedrockWorkerManager.php
iwiznia: TIL

bin/BedrockWorkerManager.php
mattabullock: Huh didn't know that was a thing.

mattabullock:
Updated

bin/BedrockWorkerManager.php
```diff
@@ -43,36 +43,30 @@
 
 // Add defaults
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
-$maxLoad = floatval($options['maxLoad'] ?? 1.0); // Max load of 1.0 by default
-$maxIterations = intval($options['maxIterations'] ?? -1); // Unlimited iterations by default
+$maxLoad = floatval(@$options['maxLoad']) ?: 1.0; // Max load of 1.0 by default
```
iwiznia: Argh, don't like the `@` can you change this to `floatval($options['maxLoad'] ?? 1.0)`?

bin/BedrockWorkerManager.php
```diff
@@ -237,15 +231,16 @@
...
                         $worker = new $workerName($bedrock, $job);
                         $childPID = getmypid();
                         try {
-                            $localDB->write("INSERT INTO localJobs (localJobID, pid, jobID, jobName, started) values ($localJobID, $childPID, {$job['jobID']}, '{$job['name']}', DATETIME('now'));");
+                            $localDB->write("INSERT INTO localJobs (pid, jobID, jobName, started) VALUES ($childPID, {$job['jobID']}, '{$job['name']}', DATETIME('now'));");
 
                             // Run the worker.  If it completes successfully, finish the job.
                             $worker->run();
 
+                            $localJobID = $localDB->getLastInsertedRowID();
```
iwiznia: 1. Move this to before `$worker->run()` just so it's closer to the insert.
2. I don't think you are using a different instance of the SQLite object for each worker, which is needed if you want `getLastInsertedRowID` to work properly.

bin/BedrockWorkerManager.php
```diff
@@ -310,35 +305,47 @@
...
     }
 
     // Calculate the speed of the last 2 batches to see if we're speeding up or slowing down
     $oldBatchTime = 0;
-    $oldBatchTimes = $localDB->read("SELECT cast(strftime('%s', ended)-strftime('%s', started) as int) from localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
-    if (is_array($oldBatchTimes)) {
-        $oldBatchTime = array_sum($oldBatchTimes)/count($oldBatchTimes);
-    }
+    $oldBatchTimes = $localDB->read("SELECT cast(strftime('%s', ended)-strftime('%s', started) AS int) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
+    $oldBatchTime = array_sum($oldBatchTimes)/count($oldBatchTimes);
```
iwiznia: NAB: can you add spaces around the `/`

iwiznia: NAB: can you add spaces around the `/`?

iwiznia: NAB: can you add spaces around the `/`?

bin/BedrockWorkerManager.php
mattabullock: I'm fairly certain it is, since it happens inside the fork and when you fork it creates a copy of all the variables from the parent.

iwiznia: But... does it create a new connection or not? I really doubt it does.

bin/BedrockWorkerManager.php
mattabullock: Ah yeah, I think this was a bad merge. with master.

bin/BedrockWorkerManager.php
mattabullock: >On Unix platforms, SQLite is sensitive to scripts that use the fork() system call. If you do have such a script, it is recommended that you close the handle prior to forking and then re-open it in the child and/or parent. For more information on this issue, see » The C language interface to the SQLite library in the section entitled Multi-Threading And SQLite.

via: http://php.net/manual/en/function.sqlite-open.php

I'll look into doing that instead.

mattabullock:
Updated with closing sqlite before forking

bin/BedrockWorkerManager.php
```diff
@@ -181,6 +201,10 @@
                 $logger->info("Forking and running a worker.", [
                     'workerFileName' => $workerFilename,
                 ]);
+
+                // Close DB connection before forking.
+                $localDB->close();
```
cead22: The handle is used both by parent and child threads

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,44 @@
...
+$minSafeJobs = intval($options['minSafeJobs']) ?: 10;  // The minimum number of jobs before we start paying attention
+$maxSafeTime = intval($options['maxSafeTime']) ?: 0; // The maximum job time before we start paying attention
+$debugThrottle = $options['debugThrottle'] ?: false; // Set to true to maintain a debug history
+$disableLoadHandler = $options['disableLoadHandler'] ?: false;
+$target = $minSafeJobs;
+
+$localDB = new LocalDB($pathToDB);
+
+// Set up the database for the AIMD load handler.
+if (!$disableLoadHandler) {
```
fnwbr: NAB: A positive variable name (as `$loadHandlerEnabled` or `$enableLoadHandler`) would be easier to read imho; also according to https://github.com/Expensify/Style-Guide/blob/master/general.md#simplicity:

> Use positive boolean variables names. Eg: instead of $shouldIgnoreTriggers use $includeTriggers and negate it when using the variable.

bin/BedrockWorkerManager.php
```diff
@@ -211,12 +235,18 @@
...
                         $worker = new $workerName($bedrock, $job);
+                        $childPID = getmypid();
                         try {
+                            $localDB->write("INSERT INTO localJobs (pid, jobID, jobName, started) VALUES ($childPID, {$job['jobID']}, '{$job['name']}', DATETIME('now'));");
+
                             // Run the worker.  If it completes successfully, finish the job.
                             $worker->run();
 
+                            $localJobID = $localDB->getLastInsertedRowID();
+                            $localDB->write("UPDATE localJobs SET ended=DATETIME('now') WHERE localJobID=$localJobID;");
```
cead22: This smells like race condition
- W1 is inserted into the DB with localID=1
- W1 starts running
- W2 is inserted into the DB with localID=2
- W2 starts running
- W1 finishes and writes `ended` in the row for W2

Can you make sure to test this scenario and verify it works as designed?

src/LocalDB.php
```diff
@@ -0,0 +1,94 @@
...
+
+        $this->handle->query($query);
+    }
+
+    /**
+     * Gets last inserted row.
+     *
+     * @return int
+     */
+    public function getLastInsertedRowID()
```
fnwbr: `: int` return type

bin/BedrockWorkerManager.php
cead22: Why are we using an auto increment as a key instead of (jobID,PID)?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+ * Determines whether or not we call GetJob and try to start a new job
+ *
+ * @param LocalDB          $localDB
+ * @param int              $target
+ * @param int              $maxSafeTime
+ * @param int              $minSafeJobs
+ * @param bool             $debugThrottle
+ * @param Expensify\Logger $logger
+ *
+ * @return array
```
cead22: Explain the return values since it's unusual

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+ */
+function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $debugThrottle, Expensify\Logger $logger) : array
+{
+    // Have we hit our target job count?
+    $query = 'SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;';
```
cead22: I think you should count the primary key (though it's possible that the DB will already use the index to count even with the query as it is, but you should `explain` it and see)

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
 $status = null;
 pcntl_wait($status);
 $logger->info('Stopped BedrockWorkerManager');
+
+/**
+ * Determines whether or not we call GetJob and try to start a new job
+ *
+ * @param LocalDB          $localDB
+ * @param int              $target
+ * @param int              $maxSafeTime
```
cead22: You should describe these here

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+ * @param bool             $debugThrottle
+ * @param Expensify\Logger $logger
+ *
+ * @return array
+ */
+function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $debugThrottle, Expensify\Logger $logger) : array
+{
+    // Have we hit our target job count?
+    $query = 'SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;';
+    $numActive = $localDB->read($query)[0];
```
cead22: You should put the whole query in the call like below since it's simple and it'd be consistent (if they were long I'd be for putting them both on variables like here)

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+
+    if ($numActive < $target) {
+        // Still in a safe zone, don't worry about load
+        $logger->info("Safe to start new job", ['numActive' => $numActive, 'target' => $target]);
+        return [true, $target];
+    }
+
+    // We're at or over our target; do we have enough data to evaluate the speed?
+    $numFinished = $localDB->read('SELECT COUNT(*) FROM localJobs WHERE ended IS NOT NULL;')[0];
+    if ($numFinished < $target*2) {
```
cead22: Use spaces around operators

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+        // Wait until we finish at least two batches of our target so we can evaluate its speed,
+        // before expanding the batch.
+        $logger->info("Haven't finished two batches of target, not queuing job", ['numActive' => $numActive, 'target' => $target]);
+        return [false, $target];
+    }
+
+    // Calculate the speed of the last 2 batches to see if we're speeding up or slowing down
+    $oldBatchTime = 0;
+    $oldBatchTimes = $localDB->read("SELECT cast(strftime('%s', ended)-strftime('%s', started) AS int) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
```
cead22: NULL is not the default value of ended so I think you should either:
- Set it to null when inserting a row (preferred)
- Check here that is not null or empty string
- Add a comment stating that by default when inserting a row it'll be set to null, above the line with the INSERT

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+    $oldBatchTime = array_sum($oldBatchTimes) / count($oldBatchTimes);
```
cead22: $oldBatchAverageTime, nab

bin/BedrockWorkerManager.php
cead22: Use space around operators

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+
+    $newBatchTime = $localDB->read("SELECT sum(cast(strftime('%s', ended)-strftime('%s', started) as int))/count(*) from localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target;")[0];
```
cead22: Use space around operators and capitalize SQL words

bin/BedrockWorkerManager.php
cead22: - Use space around operators
- Use proper capitalization for table and column names

bin/BedrockWorkerManager.php
cead22: Shouldn't we be skipping all the queries when `!$debugThrottle`?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +308,67 @@
...
+        // Also, blow away any data from more than two batches ago, because we don't
+        // look farther back than that and don't want to accumulate data infinitely.  However,
+        // this is very useful data to keep while debugging to analyze our throttle behavior.
+        if (!$debugThrottle) {
+            $localDB->write("DELETE FROM localjobs WHERE localjobid IN (SELECT localjobid FROM localjobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT -1 OFFSET $target*2);");
+        }
+
+        // Authorize one more job given that we've just increased the target by one.
+        return [true, $target];
+    } else if ($newBatchTime < $maxSafeTime || $newBatchTime < 1.5 * $oldBatchTime) {
```
cead22: Why do we cut back our target when `$newBatchTime < $maxSafeTime`?

bin/BedrockWorkerManager.php
cead22: I see now that the sqlite wrapper handles opening the connection when it's been closed, but it's very weird to have this logic inside and outside of the handler

bin/BedrockWorkerManager.php
mattabullock: Part 1: No because `lastInsertedRowID()` in PHP goes on the last inserted row ID from that connection. ([source](http://php.net/manual/en/sqlite3.lastinsertrowid.php)) That's why we're closing the connection before the fork and then re-opening.

Part 2: Because AFAIK, both jobID and PID are not unique, so the combination could be not unique, and could screw with it.

bin/BedrockWorkerManager.php
mattabullock: No, the point of `debugThrottle` is to not delete from the localJobs DB so we are able to inspect it afterwards.

```sqlite> .schema localjobs
CREATE TABLE localJobs (
        pid integer NOT NULL,
        jobID integer NOT NULL,
        jobName text NOT NULL,
        started text NOT NULL,
        ended text
    );
CREATE UNIQUE INDEX NewIndex0 ON localJobs (localJobID ASC);
sqlite> explain query plan SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;
0|0|0|SCAN TABLE localJobs
sqlite> explain query plan SELECT COUNT(localJobID) FROM localJobs WHERE ended IS NULL;
0|0|0|SCAN TABLE localJobs```

File: bin/BedrockWorkerManager.php (hunk 30 above)
Comment (mattabullock) [inline bin/BedrockWorkerManager.php]: Isn't NULL the default default value?

File: src/LocalDB.php (hunk 24 above)
Comment (mattabullock) [inline src/LocalDB.php]: It can return null so I think you can't do that (someone correct me if I'm wrong).

Comment (fnwbr) [inline src/LocalDB.php]: Not 100% sure, but http://php.net/manual/en/sqlite3.lastinsertrowid.php says it's only `int`
If it's still also `null` you need to update the PHPdoc

Comment (mattabullock) [inline src/LocalDB.php]: The function here returns null if handle isn't set (just as a precaution), not `lastInsertRowID()`. I'll update the doc.

File: bin/BedrockWorkerManager.php (hunk 26 above)
Comment (cead22) [inline bin/BedrockWorkerManager.php]:
```
sqlite> CREATE TABLE localJobs (
   ...>         localJobID integer PRIMARY KEY AUTOINCREMENT NOT NULL,
   ...>         pid integer NOT NULL,
   ...>         jobID integer NOT NULL,
   ...>         jobName text NOT NULL,
   ...>         started text NOT NULL,
   ...>         ended text
   ...>     );
sqlite> explain query plan select count(1) from localjobs;
selectid    order       from        detail
----------  ----------  ----------  --------------------
0           0           0           SCAN TABLE localjobs
sqlite> create index idx on localJobs (localJobID);
sqlite> explain query plan select count(1) from localjobs;
selectid    order       from        detail
----------  ----------  ----------  ---------------------------------------------
0           0           0           SCAN TABLE localjobs USING COVERING INDEX idx
```

Comment (cead22) [inline bin/BedrockWorkerManager.php]: >No, the point of debugThrottle is to not delete from the localJobs DB so we are able to inspect it afterwards.
Noted, I got confused

File: bin/BedrockWorkerManager.php (hunk 22 above)
Comment (iwiznia) [inline bin/BedrockWorkerManager.php]: YES! +100! Death to double negations.

File: bin/BedrockWorkerManager.php [hunk 34]
```diff
@@ -29,22 +30,44 @@
...
+$maxSafeTime = intval($options['maxSafeTime']) ?: 0; // The maximum job time before we start paying attention
+$debugThrottle = $options['debugThrottle'] ?: false; // Set to true to maintain a debug history
+$disableLoadHandler = $options['disableLoadHandler'] ?: false;
+$target = $minSafeJobs;
+
+$localDB = new LocalDB($pathToDB);
+
+// Set up the database for the AIMD load handler.
+if (!$disableLoadHandler) {
+    $query = "CREATE TABLE IF NOT EXISTS localJobs (
```
iwiznia: NAB: `'`

bin/BedrockWorkerManager.php
iwiznia: Yeah... what I had in mind was really having 2 `LocalDB` objects... One for the master thread and one for each forked thread. I think it will be simpler to reason about it if we do that.

bin/BedrockWorkerManager.php
iwiznia: I think I commented this before, but I think you should move `$localJobID = $localDB->getLastInsertedRowID();` to right after the insert.

iwiznia: OH! I also realized that you are missing updating the job in the `catch` for when a worker throws.

bin/BedrockWorkerManager.php
iwiznia: I think it is.

iwiznia:
Given that [this](https://github.com/Expensify/Bedrock-PHP/pull/28#discussion_r128667560), please add tests for processing jobs that fail.

mattabullock:
Updated with fix for failed jobs and new testing for failed jobs.

mattabullock:
Updated with fix for failed jobs and new testing for failed jobs.

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,45 @@
...
 }
 
 // Add defaults
 $jobName = $options['jobName'] ?? '*'; // Process all jobs by default
 $maxLoad = floatval($options['maxLoad'] ?? 1.0); // Max load of 1.0 by default
 $maxIterations = intval($options['maxIterations'] ?? -1); // Unlimited iterations by default
-$maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads'] ?? 5); // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
+
+$pathToDB = $options['localJobsDBPath'] ?? '/tmp/localJobsDB.sql';
+$minSafeJobs = intval($options['minSafeJobs'] ?: 10);  // The minimum number of jobs before we start paying attention
```
cead22: I think these two lines will still trigger php notices if the indices are not defined

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+
+        return true;
+    }
+
+    // Have we hit our target job count?
+    $numActive = $localDB->read('SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;')[0];
+
+    if ($numActive < $target) {
+        // Still in a safe zone, don't worry about load
+        $logger->info("Safe to start new job", ['numActive' => $numActive, 'target' => $target]);
```
cead22: Mixed double and single quotes (here and below)

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+        // this is very useful data to keep while debugging to analyze our throttle behavior.
+        if (!$debugThrottle) {
+            $localDB->write("DELETE FROM localJobs WHERE localJobID IN (SELECT localJobID FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT -1 OFFSET $target * 2);");
+        }
+
+        // Authorize one more job given that we've just increased the target by one.
+        return [true, $target];
+    } else if ($newBatchAverageTime > $maxSafeTime || $newBatchAverageTime < 1.5 * $oldBatchAverageTime) {
+        // Things seem to be slowing down, pull our target back a lot
+        $target = max(floor($target/2), $minSafeJobs);
```
cead22: Style

bin/BedrockWorkerManager.php
iwiznia: Nope, that's not true, even the variable could be undefined and this would be fine.

cead22:
```
>>> $a = [];
=> []
>>> $a['carlos'] ?: 5;
PHP error:  Undefined index: carlos on line 1
```

cead22:
```
>>> $a['carlos'] ?? 5;
=> 5
```
(no error or notice)

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+ * @param int              $maxSafeTime   Maximum safe average time for a batch of jobs before it cuts back.
+ * @param int              $minSafeJobs   A number of jobs that will always be safe to run.
+ * @param bool             $debugThrottle If true, doesn't delete jobs from localDB
+ * @param Expensify\Logger $logger
+ *
+ * @return array First value is whether or not it is safe to start a new job, second is an updated $target value.
+ */
+function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $enableLoadHandler, bool $debugThrottle, Expensify\Logger $logger) : array
+{
+    if (!$enableLoadHandler) {
```
quinthar: NAB: Comment here please -- why do we have this flag anyway?  Is this just for testing, and do you intend to remove it when proven?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+        $logger->info("Load handler not enabled");
+
```
quinthar: NAB: extra line

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+        return true;
+    }
+
+    // Have we hit our target job count?
+    $numActive = $localDB->read('SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;')[0];
+
```
quinthar: NAB: extra line

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+        // before expanding the batch.
+        $logger->info("Haven't finished two batches of target, not queuing job", ['numActive' => $numActive, 'target' => $target]);
+        return [false, $target];
+    }
+
+    // Calculate the speed of the last 2 batches to see if we're speeding up or slowing down
+    $oldBatchAverageTime = 0;
+    $oldBatchTimes = $localDB->read("SELECT CAST(STRFTIME('%s', ended) - STRFTIME('%s', started) AS int) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
+    $oldBatchAverageTime = array_sum($oldBatchTimes) / count($oldBatchTimes);
+
```
quinthar: NAB: extraneous line

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+    $newBatchAverageTime = $localDB->read("SELECT SUM(CAST(STRFTIME('%s', ended) - strftime('%s', started) AS int)) / count(*) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target;")[0];
+    if (($newBatchAverageTime < $maxSafeTime || $newBatchAverageTime < 1.1 * $oldBatchAverageTime) && $numActive <= $target) {
+        $logger->info("Increasing target", ['newBatchAverageTime' => $newBatchAverageTime, 'oldBatchAverageTime' => $oldBatchAverageTime, 'numActive' => $numActive, 'target' => $target]);
```
quinthar: NAB: Can you put this under the comment so we can stick with the consistent block commenting pattern?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +315,73 @@
...
+
+        // Authorize one more job given that we've just increased the target by one.
+        return [true, $target];
+    } else if ($newBatchAverageTime > $maxSafeTime || $newBatchAverageTime < 1.5 * $oldBatchAverageTime) {
+        // Things seem to be slowing down, pull our target back a lot
+        $target = max(floor($target/2), $minSafeJobs);
+        $logger->info("Jobs are slowing down, halving the target", ['newBatchAverageTime' => $newBatchAverageTime, 'oldBatchAverageTime' => $oldBatchAverageTime, 'numActive' => $numActive, 'target' => $target]);
+    }
+
+    return [false, $target]; // Don't authorize BWM to call GetJobs
```
quinthar: NAB: Can you put the comment on the line above for consistent commenting style?

src/LocalDB.php
```diff
@@ -0,0 +1,92 @@
...
+    /** @var SQLite3 $handle */
+    private $handle;
+
+    /** @var string $location */
+    private $location;
+
+    /**
+     * Creates a localDB object and sets the file location.
+     *
+     * @param string $location
```
quinthar: NAB: I don't think we need this given the PHP7 type hint.

src/LocalDB.php
```diff
@@ -0,0 +1,92 @@
...
+        if (isset($this->handle)) {
+            $this->handle->close();
+            unset($this->handle);
+        }
+    }
+
+    /**
+     * Runs a read query on a local database.
+     *
+     * @param string $query
```
quinthar: NAB: I don't think we need this given the PHP7 type hint.

src/LocalDB.php
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @return array|null
+     */
+    public function read(string $query)
```
quinthar: NAB: can you provide a type hint for the return value, and remove the extraneous PHPdoc?

src/LocalDB.php
```diff
@@ -0,0 +1,92 @@
...
+            $returnValue = $result->fetchArray(SQLITE3_NUM);
+        }
+
+        return $returnValue ?? null;
+    }
+
+    /**
+     * Runs a write query on a local database.
+     *
+     * @param string $query
```
quinthar: NAB: I don't think we need this given the PHP7 type hint.

src/LocalDB.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function write(string $query)
+    {
+        $this->handle->query($query);
+    }
+
+    /**
+     * Gets last inserted row.
+     *
+     * @return int|null
```
quinthar: NAB: Can you use a type hint for the return value to avoid this extraneous PHPdoc line?

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,45 @@
...
 $maxIterations = intval($options['maxIterations'] ?? -1); // Unlimited iterations by default
-$maxNumberWorkerThreads = intval($options['maxNumberWorkerThreads'] ?? 5); // Max number of worker threads for temporary load handling. TODO: Remove this in favor of better solution
+
+$pathToDB = $options['localJobsDBPath'] ?? '/tmp/localJobsDB.sql';
+$minSafeJobs = intval($options['minSafeJobs'] ?: 10);  // The minimum number of jobs before we start paying attention
+$maxSafeTime = intval($options['maxSafeTime'] ?: 0); // The maximum job time before we start paying attention
+$debugThrottle = isset($options['debugThrottle']); // Set to true to maintain a debug history
+$enableLoadHandler = isset($options['enableLoadHandler']); // Enables the AIMD load handler
+$target = $minSafeJobs;
+
```
quinthar: Actually, can you just move the database initialization code under the `// Set up the database for the AIMD load handler.` comment?

bin/BedrockWorkerManager.php
iwiznia: 🤦‍♂️ I thought it was `??`... you are right.

src/LocalDB.php
iwiznia: PHP7.0 doesn't support return value type annotation for nullable types (a return type that can also be null), which is indeed supported by PHP7.1, so we'll need to wait till we migrate to that version for doing that.

src/LocalDB.php
iwiznia: Good observations, will open a new issue to discuss leaving these out if it's type annotated and there's no description for the param needed.

quinthar:
What's the latest on this?

cead22:
Matt B is on vacay ;) He's back next week according to gcal

bin/BedrockWorkerManager.php
mattabullock: It's mostly for testing. If we find the results are bad or something is broken we can disable it quickly without doing a new deploy. It could also help with debugging in the future. I can add a `TODO: remove after testing` if you think that's appropriate.

src/LocalDB.php
mattabullock: Can't use a type hint for return value if it can return null unfortunately.

mattabullock:
Updated

iwiznia:
Seems like this is ready to go, I don't see any remaining comments left to address. @quinthar can yo re-review and approve/merge if all is good?

iwiznia:
Correction, *do not merge*, just approve. We need to create the new composer version before merging.

mattabullock:
We can merge again, I updated the version.

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,43 @@
...
+$minSafeJobs = intval($options['minSafeJobs'] ?? 10);  // The minimum number of jobs before we start paying attention
+$maxSafeTime = intval($options['maxSafeTime'] ?? 0); // The maximum job time before we start paying attention
+$debugThrottle = isset($options['debugThrottle']); // Set to true to maintain a debug history
+$enableLoadHandler = isset($options['enableLoadHandler']); // Enables the AIMD load handler
+$target = $minSafeJobs;
+
+// Set up the database for the AIMD load handler.
+$localDB = new LocalDB($pathToDB);
+$localDB->open();
+
```
quinthar: NAB: Please remove blank line

bin/BedrockWorkerManager.php
```diff
@@ -211,8 +233,14 @@
...
                     // worker without needing to restart the parent.
                     include_once $workerFilename;
-                    $stats->benchmark('bedrockJob.finish.'.$job['name'], function () use ($workerName, $bedrock, $jobs, $job, $extraParams, $logger) {
+                    $stats->benchmark('bedrockJob.finish.'.$job['name'], function () use ($workerName, $bedrock, $jobs, $job, $extraParams, $logger, $localDB) {
                         $worker = new $workerName($bedrock, $job);
+                        $childPID = getmypid();
+
+                        // Open the DB connection after the fork in the child process.
+                        $localDB->open();
+                        $localDB->write("INSERT INTO localJobs (pid, jobID, jobName, started) VALUES ($childPID, {$job['jobID']}, '{$job['name']}', DATETIME('now'));");
```
quinthar: Out of curiosity, why "...{$job['jobID']}..." instead of just "...$jobID[jobID]..."?

bin/BedrockWorkerManager.php
```diff
@@ -278,3 +311,72 @@
...
+        $logger->info("Haven't finished two batches of target, not queuing job", ["numActive" => $numActive, "target" => $target]);
+        return [false, $target];
+    }
+
+    // Calculate the speed of the last 2 batches to see if we're speeding up or slowing down
+    $oldBatchAverageTime = 0;
+    $oldBatchTimes = $localDB->read("SELECT CAST(STRFTIME('%s', ended) - STRFTIME('%s', started) AS int) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
+    $oldBatchAverageTime = array_sum($oldBatchTimes) / count($oldBatchTimes);
+    $newBatchAverageTime = $localDB->read("SELECT SUM(CAST(STRFTIME('%s', ended) - strftime('%s', started) AS int)) / count(*) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target;")[0];
+    if (($newBatchAverageTime < $maxSafeTime || $newBatchAverageTime < 1.1 * $oldBatchAverageTime) && $numActive <= $target) {
```
quinthar: Why do we have `$numActive <= $target` here?  We already verify `$numActive >= $target` on line 337.  So this basically is saying "don't increase if we're at our target" -- but isn't that precisely when we *do* want to increase?

bin/BedrockWorkerManager.php
```diff
@@ -29,22 +30,43 @@
...
+$debugThrottle = isset($options['debugThrottle']); // Set to true to maintain a debug history
+$enableLoadHandler = isset($options['enableLoadHandler']); // Enables the AIMD load handler
+$target = $minSafeJobs;
+
+// Set up the database for the AIMD load handler.
+$localDB = new LocalDB($pathToDB);
+$localDB->open();
+
+if ($enableLoadHandler) {
+    $query = 'CREATE TABLE IF NOT EXISTS localJobs (
```
quinthar: We should probably open this up in WAL mode such that everybody can read it even while others are trying to write.  Also, should probably add some indexes to make those reads fast.

bin/BedrockWorkerManager.php
mattabullock: The reason I put this condition in is because I was worried about situations where we just cut the load in half because jobs were going slower and then the target increasing before the actual load drops down to 20. Example:

1. `$target` is at 20
2. Jobs slow down
3. `$target` is cut down to 10 but currently there are 20 jobs running
4. `$numActive` > `$target` and `$newBatchAverageTime < 1.1 * $oldBatchAverageTime` so the target increases before the load drops to the new max.

I don't think this is probably the worst situation, but it kind of goes against the AIMD algorithm (though we've kind of strayed from the AIMD algorithm here).

quinthar: Hm, I'm not sure where we were discussing this previously, so moving the conversation here. 
 However,  you mane an interesting point about adding that condition in there to avoid "double-reducing".  However, I'd suggest a different approach:

1) First thing, before even checking whether we are over or under target, *if there is adequate batch data* evaluate the relative speed of the last two batches:
    . If we're speeding up *and at the target*, increment the target
    . If we're slowing down *regardless of how many jobs are running*, halve the target *and delete the batch data* (or mark it in such a way that prevents it from being considered, so we can keep it for debugging reasons).  This will prevent a "double reduction".

2) Then *after* the target has been updated, authorize or decline a new job based on how many are running relative to the new target.

Currently you skip the target adjustment code entirely if under the target, but in a scenario where the target is set unobtainably high, we never reduce it.

quinthar: That said, given locking concerns, perhaps we just remove the debugging ability to retain past data, thereby keeping the table very small.  We can monitor its performance adequately from the logs, so long as we log whenever updating the target.

PR #35, opened 2017-08-15
Add potential performance improvements for BWM load handler

mattabullock:
This PR introduces a few potential performance improvements for the BWM load handler.

1. Only open/close the connection if the load handler is enabled.
2. Enable WAL mode on the database. This is persistent so should only have to happen on DB creation.
3. Create an index on the `localJobID` if one doesn't exist already.

# Tests
1. Delete localJobsDB.sql if it exists.
2. Queue test jobs
3. Run BWM with load handler enabled.
4. Confirm that there are no errors.
5. Open up `/tmp/localJobsDB.sql` and run `PRAGMA journal_mode;` and confirm the output is `wal`.
6. Run `.schema localJobs` and confirm the output is
```
CREATE TABLE localJobs (
        pid integer NOT NULL,
        jobID integer NOT NULL,
        jobName text NOT NULL,
        started text NOT NULL,
        ended text
    );
CREATE INDEX idx ON localJobs (localJobID);
```

**Please** review/approve it then let me package it before merging.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -64,7 +64,7 @@ if ($enableLoadHandler) {
         started text NOT NULL,
         ended text
     );
-    CREATE INDEX IF NOT EXISTS idx ON localJobs (localJobID);
+    CREATE INDEX IF NOT EXISTS localJobsLocalJobID ON localJobs (localJobID);
     PRAGMA journal_mode = WAL;';
     $localDB->write($query);
 }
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -53,18 +53,19 @@
...
     $query = 'CREATE TABLE IF NOT EXISTS localJobs (
         localJobID integer PRIMARY KEY AUTOINCREMENT NOT NULL,
         pid integer NOT NULL,
         jobID integer NOT NULL,
         jobName text NOT NULL,
         started text NOT NULL,
         ended text
-    );';
+    );
+    CREATE INDEX IF NOT EXISTS idx ON localJobs (localJobID);
```
cead22: `idx` is a bad name, use `localJobsLocalJobID`

cead22:
> Please review/approve it then let me package it before merging.

Added a HOLD cause I don't trust myself to remember this

mattabullock:
Updated

mattabullock:
Ready for merge!

PR #38, opened 2017-08-30
Catch localJobsDB sqlite exception and retry immediately

mattabullock:
The local SQLite DB is having locking issues so when it runs into an error we'll log the error and retry.

Related to https://github.com/Expensify/Expensify/issues/60379#issuecomment-325038663

# Tests
1. Queue a million jobs
2. Run bwm with enableLoadHandler
3. Wait forever
4. `grep 'database is locked' syslog` and confirm that it doesn't appear
5. `grep 'Query failed, retrying' syslog` and confirm that it does appear
6. Run bwm without enableLoadHandler
7. Confirm everything runs with no errors.

### DIFF
--- a/src/LocalDB.php
+++ b/src/LocalDB.php
@@ -66,7 +66,12 @@ class LocalDB
                 $result = $this->handle->query($query);
                 break;
             } catch (Exception $e) {
-                $this->logger->info("Query failed, retrying", ['query' => $query, 'error' => json_encode($e)]);
+                if ($e->getMessage() === 'database is locked') {
+                    $this->logger->info("Query failed, retrying", ['query' => $query, 'error' => $e->getMessage()]);
+                } else {
+                    $this->logger->info("Query failed, not retrying", ['query' => $query, 'error' => $e->getMessage()]);
+                    throw $e;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/LocalDB.php
```diff
@@ -53,7 +60,15 @@ public function close()
      */
     public function read(string $query)
     {
-        $result = $this->handle->query($query);
+        $result = null;
+        while(true) {
+            try {
+                $result = $this->handle->query($query);
+                break;
+            } catch (Exception $e) {
```
iwiznia: I think you should catch specific exceptions, not all of them. So either by matching by exception type or by it's string, check the exception is the one we want and only retry in those cases.

src/LocalDB.php
```diff
@@ -53,7 +60,15 @@ public function close()
...
+                $this->logger->info("Query failed, retrying", ['query' => $query, 'error' => json_encode($e)]);
```
iwiznia: How does the `json_encode($e)` looks like in the log? I didn't know you could serialize an exception like that 😄

mattabullock:
Updated

iwiznia:
Looks good, you need to create the new version via `package.sh`, BUT! Don't do it till you merge my PR https://github.com/Expensify/Bedrock-PHP/pull/39 or we'll conflict with each other.

PR #39, opened 2017-09-01
Prevent error jobID

iwiznia:
@mattabullock please review.

Some commands don't return either name nor jobID and this fails when that happens.


mattabullock:
Any tests?

iwiznia:
Tested with the web PR.

mattabullock:
Can you link it in the issue? I know it's a trivial change, but given the recent email I figured it's better to document it even if it's trivial.

PR #41, opened 2017-09-21
Improved failover code

iwiznia:
@flodnv please review. /cc @quinthar 

This implements the new failover code we talked about in https://github.com/Expensify/Expensify/issues/54343

See all tests in the [web PR](https://github.com/Expensify/Web-Expensify/pull/18557).

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -31,7 +31,7 @@ class Client implements LoggerAwareInterface
     /**
      * The prefix to use in apcu to store the state of the hosts.
      */
-    const APCU_CACHE_PREFIX = 'bedrockHostStates-';
+    const APCU_CACHE_PREFIX = 'bedrockHostConfigs-';
 
     /**
      * @var array This is a default configuration applied to all instances of this class. They can be overriden in the
@@ -357,32 +357,32 @@ class Client implements LoggerAwareInterface
 
     private function getPossibleHosts()
     {
-        // If cached hosts aren't set, get them from the APC cache. Then, we check the configuration there with the passed
+        // We get the host configs from the APC cache. Then, we check the configuration there with the passed
         // configuration and if it's outdated (ie: it has different hosts from the one in the config), we reset it. This
         // is so that we don't keep the old cache after changing the hosts or failover configuration.
         if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING)) {
             $apcuKey = self::APCU_CACHE_PREFIX.$this->clusterName;
-            $hostConfigs = apcu_fetch($apcuKey) ?: [];
-            $this->logger->info('Bedrock\Client - APC fetch failover hosts', $hostConfigs);
+            $cachedHostConfigs = apcu_fetch($apcuKey) ?: [];
+            $this->logger->info('Bedrock\Client - APC fetch host configs', $cachedHostConfigs);
 
             // If the hosts and ports in the cache don't match the ones in the config, reset the cache.
-            $savedHostsAndPort = [];
-            foreach ($hostConfigs as $hostName => $config) {
-                $savedHostsAndPort[$hostName] = $config['port'];
+            $cachedHostsAndPorts = [];
+            foreach ($cachedHostConfigs as $hostName => $config) {
+                $cachedHostsAndPorts[$hostName] = $config['port'];
             }
-            asort($savedHostsAndPort);
-            $configHostsAndPort = [];
+            asort($cachedHostsAndPorts);
+            $uncachedHostsAndPort = [];
             foreach (array_merge($this->mainHostConfigs, $this->failoverHostConfigs) as $hostName => $config) {
-                $configHostsAndPort[$hostName] = $config['port'];
+                $uncachedHostsAndPort[$hostName] = $config['port'];
             }
-            asort($configHostsAndPort);
-            if ($savedHostsAndPort !== $configHostsAndPort) {
-                $hostConfigs = array_merge($this->mainHostConfigs, $this->failoverHostConfigs);
-                $this->logger->info('Bedrock\Client - APC init failover hosts', $hostConfigs);
-                apcu_store($apcuKey, $hostConfigs);
+            asort($uncachedHostsAndPort);
+            if ($cachedHostsAndPorts !== $uncachedHostsAndPort) {
+                $cachedHostConfigs = array_merge($this->mainHostConfigs, $this->failoverHostConfigs);
+                $this->logger->info('Bedrock\Client - APC init host configs', $cachedHostConfigs);
+                apcu_store($apcuKey, $cachedHostConfigs);
             }
         } else {
-            $hostConfigs = array_merge($this->mainHostConfigs, $this->failoverHostConfigs);
+            $cachedHostConfigs = array_merge($this->mainHostConfigs, $this->failoverHostConfigs);
         }
 
         // Get one main host and all the failovers, then remove any of them that we know already failed.
@@ -399,13 +399,13 @@ class Client implements LoggerAwareInterface
 
         $nonBlackListedHosts = [];
         foreach ($hostNames as $hostName) {
-            $blackListedUntil = $hostConfigs[$hostName]['blacklistedUntil'] ?? null;
+            $blackListedUntil = $cachedHostConfigs[$hostName]['blacklistedUntil'] ?? null;
             if (!$blackListedUntil || $blackListedUntil < time()) {
-                $nonBlackListedHosts[$hostName] = $hostConfigs[$hostName];
+                $nonBlackListedHosts[$hostName] = $cachedHostConfigs[$hostName];
             }
         }
 
-        $this->getLogger()->info('Bedrock\Client - Possible hosts', ['mainHostConfigs' => array_keys($nonBlackListedHosts)]);
+        $this->getLogger()->info('Bedrock\Client - Possible hosts', ['nonBlacklistedHosts' => array_keys($nonBlackListedHosts)]);
 
         return $nonBlackListedHosts;
     }
--- a/src/Client.php
+++ b/src/Client.php
@@ -229,12 +229,16 @@ class Client implements LoggerAwareInterface
         $numTries = 3;
         $response = null;
         $lastTryException = null;
-        while($numTries-- && !$response) {
+        $hosts = $this->getPossibleHosts();
+        $host = null;
+        while($numTries-- && !$response && count($hosts)) {
+            reset($hosts);
+            $host = key($hosts);
             try {
                 // Do the request.  This is split up into separate functions so we can
                 // profile them independently -- useful when diagnosing various network
                 // conditions.
-                $this->sendRawRequest($rawRequest);
+                $this->sendRawRequest($host, $rawRequest);
                 $response = $this->receiveRawResponse();
 
                 // Record the last error in the response as this affects how we
@@ -262,6 +266,8 @@ class Client implements LoggerAwareInterface
                     $this->logger->error("Failed to send the whole request or to receive it; not retrying", ['message' => $e->getMessage()]);
                     throw $e;
                 }
+            } finally {
+                array_shift($hosts);
             }
         }
 
@@ -285,66 +291,50 @@ class Client implements LoggerAwareInterface
     }
 
     /**
-     * Create and connect a socket (with failovers).
-     *
-     * @throws ConnectionFailure
-     */
-    private function reconnect()
-    {
-        $this->logger->debug('Opening new socket');
-
-        $hosts = $this->getPossibleHosts();
-        $socketErrorCode = 1;
-        $host = null;
-        while ($socketErrorCode !== 0) {
-            // Try to connect to the requested host
-            if ($this->socket) {
-                socket_close($this->socket);
-                $this->socket = null;
-            }
-            $this->socket = @socket_create(AF_INET, SOCK_STREAM, getprotobyname('tcp'));
-
-            // Make sure we succeed to create a socket
-            if ($this->socket === false) {
-                $socketError = socket_strerror(socket_last_error());
-                throw new ConnectionFailure("Could not connect to create socket: $socketError");
-            }
-
-            // Configure this socket and try all the possible hosts in order, till we find one that works or all of them fail
-            socket_set_option($this->socket, SOL_SOCKET, SO_SNDTIMEO, ['sec' => $this->connectionTimeout, 'usec' => 0]);
-            socket_set_option($this->socket, SOL_SOCKET, SO_RCVTIMEO, ['sec' => $this->readTimeout, 'usec' => 0]);
-
-            reset($hosts);
-            $host = key($hosts);
-            @socket_connect($this->socket, $host, $hosts[$host]['port']);
-            $socketErrorCode = socket_last_error($this->socket);
-            socket_clear_error($this->socket);
-            $this->lastHostUsed = $host;
-            if ($socketErrorCode) {
-                $socketError = socket_strerror($socketErrorCode);
-                $this->logger->info("Failed to connect to host", ['host' => $host, 'port' => $hosts[$host]['port'], 'error' => $socketError]);
-                $this->markHostAsFailed();
-                array_shift($hosts);
-            }
-
-            // Nothing we can do, we couldn't connect to any host
-            if (empty($hosts)) {
-                throw new ConnectionFailure("Could not connect to Bedrock hosts or failovers");
-            }
-        }
-    }
-
-    /**
      * Sends the request on the existing socket, if possible, else it reconnects.
      *
      * @param string $rawRequest
      *
-     * @throws ConnectionFailure
-     * @throws BedrockError
+     * @throws ConnectionFailure When the failure is before sending any data to the server
+     * @throws BedrockError When we already sent some data
      */
-    private function sendRawRequest($rawRequest)
+    private function sendRawRequest(string $host, $rawRequest)
     {
-        $this->reconnect();
+        $this->logger->debug('Opening new socket');
+        // Try to connect to the requested host
+        if ($this->socket) {
+            socket_close($this->socket);
+            $this->socket = null;
+        }
+        $this->socket = @socket_create(AF_INET, SOCK_STREAM, getprotobyname('tcp'));
+
+        // Make sure we succeed to create a socket
+        if ($this->socket === false) {
+            $socketError = socket_strerror(socket_last_error());
+            throw new ConnectionFailure("Could not connect to create socket: $socketError");
+        }
+
+        // Configure this socket and try all the possible hosts in order, till we find one that works or all of them fail
+        socket_set_option($this->socket, SOL_SOCKET, SO_SNDTIMEO, ['sec' => $this->connectionTimeout, 'usec' => 0]);
+        socket_set_option($this->socket, SOL_SOCKET, SO_RCVTIMEO, ['sec' => $this->readTimeout, 'usec' => 0]);
+
+        @socket_connect($this->socket, $host, self::$cachedHosts[$this->clusterName][$host]['port']);
+        $socketErrorCode = socket_last_error($this->socket);
+        socket_clear_error($this->socket);
+        $this->lastHostUsed = $host;
+        if ($socketErrorCode) {
+            $socketError = socket_strerror($socketErrorCode);
+            $this->logger->info("Failed to connect to host", ['host' => $host, 'port' => $hosts[$host]['port'], 'error' => $socketError]);
+            $this->markHostAsFailed();
+            array_shift($hosts);
+        }
+
+        // Nothing we can do, we couldn't connect to any host
+        if (empty($hosts)) {
+            throw new ConnectionFailure("Could not connect to Bedrock hosts or failovers");
+        }
+
+        // Send the information to the socket
         $bytesSent = socket_send($this->socket, $rawRequest, strlen($rawRequest), MSG_EOF);
 
         // Failed to send anything
@@ -419,7 +409,7 @@ class Client implements LoggerAwareInterface
      *
      * @return string The new data received.
      *
-     * @throws ConnectionFailure
+     * @throws BedrockError
      */
     private function recv()
     {
@@ -429,9 +419,9 @@ class Client implements LoggerAwareInterface
         if ($numRecv === false) {
             $errorCode = socket_last_error();
             $errorMsg = socket_strerror($errorCode);
-            throw new ConnectionFailure("Socket read failed: '$errorMsg' #$errorCode");
+            throw new BedrockError("Socket read failed: '$errorMsg' #$errorCode");
         } elseif ($numRecv <= 0) {
-            throw new ConnectionFailure("Socket read failed: no data returned");
+            throw new BedrockError("Socket read failed: no data returned");
         }
 
         return $buf;
--- a/src/Client.php
+++ b/src/Client.php
@@ -19,56 +19,45 @@ class Client implements LoggerAwareInterface
     const HEADER_FIELD_SEPARATOR = "\r\n";
 
     /**
-     * This is a list of hosts to use with it's ports and timeout. We store it here to not have to fetch it from APC on
-     * each request.
-     * @var array
+     * @var array This is a list of hosts to use with it's ports and timeout. We store it here to not have to fetch it
+     * from APC on each request.
      */
     private static $cachedHosts = [];
 
     /**
-     * This is a default configuration applied to all instances of this class. They can be overriden in the constructor.
-     * @var array
+     * @var array This is a default configuration applied to all instances of this class. They can be overriden in the
+     * constructor.
      */
     private static $defaultConfig = [];
 
     /**
-     * Name of the bedrock cluster we are talking to. If you have more than one bedrock cluster, you can pass in different
-     * names for them in order to have separate statistics collected and caches of failed servers.
-     * @var null|string
+     * @var null|string Name of the bedrock cluster we are talking to. If you have more than one bedrock cluster, you
+     * can pass in different names for them in order to have separate statistics collected and caches of failed servers.
      */
     private $clusterName = null;
 
     /**
-     * What is the last commit count of the node we talked to.
-     *
-     * This is used to ensure if we make a subsequent request to a different
-     * node in the same session, that the node waits until it is at least as
-     * "fresh" as the node we originally queried.
-     *
-     *  @var null|string
+     *  @var null|string The last commit count of the node we talked to. This is used to ensure if we make a subsequent
+     * request to a different node in the same session, that the node waits until it is at least as "fresh" as the node
+     * we originally queried.
      */
     private $commitCount = null;
 
     /**
-     * Existing socket, if any.  It's reused each time, if possible.
-     *
-     *  @var null|resource
+     *  @var null|resource Existing socket, if any.  It's reused each time, if possible.
      */
     private $socket = null;
 
     /**
-     * List of hosts to use as first choice.
-     *
-     *  @var array
+     *  @var array List of hosts to use as first choice. It will pick just one of these randomly and try it first.
      */
-    private $hosts = [];
+    private $mainHostConfigs = [];
 
     /**
-     * List of failovers we attempt if the first didn't work.
-     *
-     *  @var array
+     *  @var array List of failovers we attempt if the first didn't work. We randomize the list and try on several of
+     * them (depending on the number of retries configured).
      */
-    private $failovers = [];
+    private $failoverConfigs = [];
 
     /**
      * @var int Timeout for connecting to the server.
@@ -110,8 +99,6 @@ class Client implements LoggerAwareInterface
      */
     private $retries;
 
-    private $lastHostUsed;
-
     /**
      * Creates a reusable Bedrock instance.
      * All params are optional and values set in `configure` would be used if are not passed here.
@@ -137,8 +124,8 @@ class Client implements LoggerAwareInterface
     {
         $config = array_merge(self::$defaultConfig, $config);
         $this->clusterName = $config['clusterName'];
-        $this->hosts = $config['hosts'];
-        $this->failovers = $config['failovers'];
+        $this->mainHostConfigs = $config['hosts'];
+        $this->failoverConfigs = $config['failovers'];
         $this->connectionTimeout = $config['connectionTimeout'];
         $this->readTimeout = $config['readTimeout'];
         $this->logger = $config['logger'];
@@ -149,8 +136,8 @@ class Client implements LoggerAwareInterface
         $this->retries = $config['retries'];
 
         // Make sure we have at least one host configured
-        $this->logger->debug('Bedrock\Client - Constructed', ['clusterName' => $this->clusterName, 'hosts' => $this->hosts, 'failovers' => $this->failovers]);
-        if (empty($this->hosts)) {
+        $this->logger->debug('Bedrock\Client - Constructed', ['clusterName' => $this->clusterName, 'hosts' => $this->mainHostConfigs, 'failovers' => $this->failoverConfigs]);
+        if (empty($this->mainHostConfigs)) {
             throw new BedrockError('Failed to construct Bedrock object');
         }
     }
@@ -168,8 +155,8 @@ class Client implements LoggerAwareInterface
         // Store the configuration
         self::$defaultConfig = array_merge([
             'clusterName' => 'bedrock',
-            'hosts' => ['localhost' => ['timeout' => 0, 'port' => 8888]],
-            'failovers' => ['localhost' => ['timeout' => 0, 'port' => 8888]],
+            'hosts' => ['localhost' => ['blacklistedUntil' => 0, 'port' => 8888]],
+            'failovers' => ['localhost' => ['blacklistedUntil' => 0, 'port' => 8888]],
             'connectionTimeout' => 1,
             'readTimeout' => 300,
             'logger' => new NullLogger(),
@@ -264,11 +251,11 @@ class Client implements LoggerAwareInterface
         $rawRequest .= $body;
 
         // We try +1 of the configured amount of retries, each time on a different valid server
-        $numTries = $this->retries + 1;
+        $numTriesRemaining = $this->retries + 1;
         $response = null;
         $hosts = $this->getPossibleHosts();
         $host = null;
-        while($numTries-- && !$response && count($hosts)) {
+        while($numTriesRemaining-- && !$response && count($hosts)) {
             reset($hosts);
             $host = key($hosts);
             try {
@@ -279,18 +266,18 @@ class Client implements LoggerAwareInterface
                 $response = $this->receiveResponse();
             } catch(ConnectionFailure $e) {
                 // The error happened during connection (or before we sent any data) so we can retry it safely
-                $this->markHostAsFailed();
-                if ($numTries) {
-                    $this->logger->info('Bedrock\Client - Failed to connect or send the request; retrying', ['host' => $host, 'message' => $e->getMessage(), 'retriesLeft' => $numTries, 'exception' => $e]);
+                $this->markHostAsFailed($host);
+                if ($numTriesRemaining) {
+                    $this->logger->info('Bedrock\Client - Failed to connect or send the request; retrying', ['host' => $host, 'message' => $e->getMessage(), 'retriesLeft' => $numTriesRemaining, 'exception' => $e]);
                 } else {
                     $this->logger->error('Bedrock\Client - Failed to connect or send the request; not retrying', ['host' => $host, 'message' => $e->getMessage(), 'exception' => $e]);
                     throw $e;
                 }
             } catch(BedrockError $e) {
                 // This error happen after sending some data to the server, so we only can retry it if it is an idempotent command
-                $this->markHostAsFailed();
-                if ($numTries && ($headers['idempotent'] ?? false)) {
-                    $this->logger->info('Bedrock\Client - Failed to send the whole request or to receive it; retrying', ['host' => $host, 'message' => $e->getMessage(), 'retriesLeft' => $numTries, 'exception' => $e]);
+                $this->markHostAsFailed($host);
+                if ($numTriesRemaining && ($headers['idempotent'] ?? false)) {
+                    $this->logger->info('Bedrock\Client - Failed to send the whole request or to receive it; retrying because command is idempotent', ['host' => $host, 'message' => $e->getMessage(), 'retriesLeft' => $numTriesRemaining, 'exception' => $e]);
                 } else {
                     $this->logger->error('Bedrock\Client - Failed to send the whole request or to receive it; not retrying', ['host' => $host, 'message' => $e->getMessage(), 'exception' => $e]);
                     throw $e;
@@ -346,15 +333,13 @@ class Client implements LoggerAwareInterface
             throw new ConnectionFailure("Could not connect to create socket: $socketError");
         }
 
-        // Configure this socket and try all the possible hosts in order, till we find one that works or all of them fail
+        // Configure this socket and try to connect to it
         socket_set_option($this->socket, SOL_SOCKET, SO_SNDTIMEO, ['sec' => $this->connectionTimeout, 'usec' => 0]);
         socket_set_option($this->socket, SOL_SOCKET, SO_RCVTIMEO, ['sec' => $this->readTimeout, 'usec' => 0]);
-
         $port = self::$cachedHosts[$this->clusterName][$host]['port'];
         @socket_connect($this->socket, $host, $port);
         $socketErrorCode = socket_last_error($this->socket);
         socket_clear_error($this->socket);
-        $this->lastHostUsed = $host;
         if ($socketErrorCode) {
             $socketError = socket_strerror($socketErrorCode);
             throw new ConnectionFailure("Could not connect to Bedrock host $host:$port. Error: $socketErrorCode $socketError");
@@ -380,7 +365,9 @@ class Client implements LoggerAwareInterface
 
     private function getPossibleHosts()
     {
-        // If cached hosts aren't set, get them from the APC cache or init them from the config if they are not set or outdated
+        // If cached hosts aren't set, get them from the APC cache. Then, we check the configuration there with the passed
+        // configuration and if it's outdated (ie: it has different hosts from the one in the config), we reset it. This
+        // is so that we don't keep the old cache after changing the hosts or failover configuration.
         if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING) && empty(self::$cachedHosts[$this->clusterName])) {
             $apcuKey = 'bedrockFailoverHosts-'.$this->clusterName;
             self::$cachedHosts[$this->clusterName] = apcu_fetch($apcuKey) ?: [];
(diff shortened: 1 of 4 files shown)


src/Jobs.php
```diff
@@ -101,19 +101,6 @@ public function call($method, $headers = [], $body = '')
 
         $this->client->getStats()->counter('bedrockJob.call.response.'.$method.$responseCode);
 
-        if ($responseCode!=200 && @$response['lastTryException']) {
```
iwiznia: This was causing issues when jobs returned 404 (no jobs to process). Since it returned 200, it was trying to use the response, which obviously wasn't there. I think it's not needed with the new failover, but not 100% sure.

src/Client.php
```diff
@@ -15,37 +15,178 @@
  */
 class Client implements LoggerAwareInterface
 {
+    const HEADER_DELIMITER = "\r\n\r\n";
+    const HEADER_FIELD_SEPARATOR = "\r\n";
+
+    /**
+     * This is a list of hosts to use with it's ports and timeout. We store it here to not have to fetch it from APC on
```
quinthar: How is this different than `$this->hosts`?

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+     * @var array
+     */
+    private static $defaultConfig = [];
+
+    /**
+     * Name of the bedrock cluster we are talking to. If you have more than one bedrock cluster, you can pass in different
+     * names for them in order to have separate statistics collected and caches of failed servers.
+     * @var null|string
+     */
+    private $clusterName = null;
```
quinthar: Do we have multiple cluster names, or is this just anticipating the possibility?

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+    /**
+     * Existing socket, if any.  It's reused each time, if possible.
+     *
+     *  @var null|resource
+     */
+    private $socket = null;
+
     /**
-     * Sets the default configuration.
+     * List of hosts to use as first choice.
```
quinthar: Perhaps clarify that we will make one random selection from this list, as opposed to `failovers` (which we will keep trying until we find one that works).

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
      *
-     * @param array $config
+     *  @var array
+     */
+    private $hosts = [];
+
+    /**
+     * List of failovers we attempt if the first didn't work.
```
quinthar: Perhaps clarify that we will keep trying these in random order until we find one that works.

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+     *  @var null|resource
+     */
+    private $socket = null;
+
     /**
-     * Sets the default configuration.
+     * List of hosts to use as first choice.
      *
-     * @param array $config
+     *  @var array
```
quinthar: The variables below have the description after the variable name, but this one has it in a paragraph above the variable name.  Is there a difference?  (Sorry, I'm not super familiar with Javadocs syntax.)

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+
+    /**
+     * @var int When a host fails, it will blacklist it and not try to reuse it for up to this amount of seconds.
+     */
+    private $maxBlackListTimeout;
+
+    /**
+     * @var int The length of each packet read from the socket.
+     */
+    private $packetLength;
```
quinthar: Why is this configurable -- can't we just use a constant?

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+
+    /**
+     * @var int The number of times to retry on a different server if the first one fails.
+     */
+    private $retries;
```
quinthar: Again, let's not make this configurable, this feels like a good constant.

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+
+    private $lastHostUsed;
```
quinthar: Missing JavaDocs.

src/Client.php
```diff
@@ -57,68 +198,21 @@ public static function getLogger()
...
-     */
-    public static function getStats()
-    {
-        if (is_string(self::$config['stats'])) {
-            self::$config['stats'] = new self::$config['stats'];
-        }
+        $this->logger = $logger;
 
-        return self::$config['stats'];
+        return null;
```
quinthar: why does this return anything?

src/Client.php
quinthar: Oh, we do!  authrock and webrock.

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
+        // whole thing, else there's a problem.
+        if ($bytesSent != strlen($rawRequest)) {
+            $this->logger->info('Bedrock\Client - Could not send the whole request', ['bytesSent' => $bytesSent, 'expected' => strlen($rawRequest)]);
+            throw new BedrockError("Sent partial request to bedrock host $host:$port");
+        }
+    }
+
+    private function getPossibleHosts()
+    {
+        // If cached hosts aren't set, get them from the APC cache or init them from the config if they are not set or outdated
```
quinthar: Hm, I really can't figure out what this is doing.  Can you expand this comment with more context?

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
+            if ($savedHostsAndPort !== $configHostsAndPort) {
+                self::$cachedHosts[$this->clusterName] = array_merge($this->hosts, $this->failovers);
+                $this->logger->info('Bedrock\Client - APC init failover hosts', self::$cachedHosts[$this->clusterName]);
+                apcu_store($apcuKey, self::$cachedHosts[$this->clusterName]);
+            }
+        } else {
+            self::$cachedHosts[$this->clusterName] = array_merge($this->hosts, $this->failovers);
         }
+
+        // Get one main host and all the failovers, then remove any of them that we know already failed.
```
quinthar: This is a great comment!  Explains it succinctly and perfectly.

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
+        // Assemble the list of servers we'll try, in order.  First, pick one of the main hosts. We pick randomly
+        // because we want to equally balance each server across all of its local databases. This allows us to have an
+        // unequal number of servers and databases in a given datacenter. Also, we only pick one (versus trying both)
+        // because if our first attempt fails we want to equally balance across *all* databases -- including the remote
+        // ones. Otherwise if a database node goes down, the other databases in the same datacenter would get more load
+        // (whereas this approach ensures the load is spread evenly across all).
+        $failovers = array_keys($this->failovers);
```
quinthar: Yuck, can you use a different variable name?  I'd suggest renaming both, so it's:

`$failoverHostnames = array_keys($this->failoverHostConfigs);`

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
+        shuffle($failovers);
+        $mainHost = array_rand($this->hosts);
+        $hosts = array_merge([$mainHost], $failovers);
```
quinthar: I'm not sure how this works.   If the variable names were more descriptive, I think it would be:

```
$mainHostConfig = array_rand($this->mainHostConfigs);
$mainHostConfigArrayWTF = [$mainHostConfig];
$hostWTF = array_merge($mainHostConfigArrayWTF, $failoverHostNames];
```

It's merging two totally different kinds of objects -- an array containing a single config, with an array of host names.  How does this work?  Or am I completely misunderstanding something?

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
+
+        $nonBlackListedHosts = [];
+        foreach ($hosts as $host) {
```
quinthar: Again some naming would make this more clear, like:

```
foreach ($hostName as $hostNames) {
```

But I don't think `$hosts` contains what you think it does.  Or, rather, I think it probably always skips over the "main host" because it's not a string?

src/Client.php
```diff
@@ -475,4 +571,14 @@ private static function toUTF8($str)
 
         return iconv($encoding, 'UTF-8//IGNORE', $str);
     }
+
+    private function markHostAsFailed()
+    {
+        $time = time() + rand(1, $this->maxBlackListTimeout);
+        self::$cachedHosts[$this->clusterName][$this->lastHostUsed]['timeout'] = $time;
+        if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING) && empty(self::$cachedHosts[$this->clusterName])) {
+            apcu_store('bedrockFailoverHosts-'.$this->clusterName, self::$cachedHosts[$this->clusterName]);
```
quinthar: So we only load this once per PHP instance?  So BWM will load it once on its initial start, but then never learn from any connection failures discovered by PHP-CGI instances on the same server?  If this is for performance reasons, do you know that this is actually a problem, or is this just a premature optimization?

src/Client.php
```diff
@@ -168,43 +263,55 @@ public function call($method, $headers = [], $body = '')
...
+                    $this->logger->info('Bedrock\Client - Failed to connect or send the request; retrying', ['host' => $host, 'message' => $e->getMessage(), 'retriesLeft' => $numTries, 'exception' => $e]);
                 } else {
-                    $this->getLogger()->error("Failed to connect, send, or receive request; not retrying", ['message' => $e->getMessage()]);
+                    $this->logger->error('Bedrock\Client - Failed to connect or send the request; not retrying', ['host' => $host, 'message' => $e->getMessage(), 'exception' => $e]);
                     throw $e;
                 }
+            } catch(BedrockError $e) {
+                // This error happen after sending some data to the server, so we only can retry it if it is an idempotent command
+                $this->markHostAsFailed();
+                if ($numTries && ($headers['idempotent'] ?? false)) {
```
quinthar: Perhaps rename `$numTries` to `$numTriesRemaining` to be more clear

src/Client.php
```diff
@@ -168,43 +263,55 @@ public function call($method, $headers = [], $body = '')
...
+                    $this->logger->info('Bedrock\Client - Failed to send the whole request or to receive it; retrying', ['host' => $host, 'message' => $e->getMessage(), 'retriesLeft' => $numTries, 'exception' => $e]);
```
quinthar: Perhaps mention in here that we're retrying because it's idempotent

src/Client.php
```diff
@@ -168,43 +263,55 @@ public function call($method, $headers = [], $body = '')
...
+                } else {
+                    $this->logger->error('Bedrock\Client - Failed to send the whole request or to receive it; not retrying', ['host' => $host, 'message' => $e->getMessage(), 'exception' => $e]);
+                    throw $e;
+                }
+            } finally {
+                array_shift($hosts);
```
quinthar: Can you put the `markHostAsFailed()` in this `finally` section?

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
-            if (!(@socket_connect($this->socket, $this->failoverHost, $this->failoverPort))) {
-                // Connection to both the main host and the failover failed!
-                $socketError = socket_strerror(socket_last_error($this->socket));
-                throw new ConnectionFailure("Could not connect to Bedrock primary host ({$this->host}) or failover ({$this->failoverHost}): $socketError");
+
+        $port = self::$cachedHosts[$this->clusterName][$host]['port'];
+        @socket_connect($this->socket, $host, $port);
+        $socketErrorCode = socket_last_error($this->socket);
+        socket_clear_error($this->socket);
+        $this->lastHostUsed = $host;
```
quinthar: I think `$lastHostUsed` is unnecessary.  The only time it's used is in `markHostAsInvalid()`, but that is called from a context that already has `$host` set to the host being tried -- so just remove `$lastHostUsed` and pass `$host` into `markHostAsInvalid()`.

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
 {
+    const HEADER_DELIMITER = "\r\n\r\n";
+    const HEADER_FIELD_SEPARATOR = "\r\n";
+
+    /**
+     * This is a list of hosts to use with it's ports and timeout. We store it here to not have to fetch it from APC on
+     * each request.
+     * @var array
+     */
+    private static $cachedHosts = [];
```
quinthar: So far as I can tell, this is totally unnecessary and a premature optimization that creates bad edge cases.  I'd suggest removing this entirely.

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+    private $socket = null;
+
     /**
-     * Sets the default configuration.
+     * List of hosts to use as first choice.
      *
-     * @param array $config
+     *  @var array
+     */
+    private $hosts = [];
```
quinthar: I'd suggest renaming this to `$mainHostConfigs`, to clarify that it's a "main" host (and not a failover host), and a full configuration of that host (not just a hostname).  (This is confusing later.)

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
+
+    /**
+     * List of failovers we attempt if the first didn't work.
+     *
+     *  @var array
+     */
+    private $failovers = [];
```
quinthar: I'd suggest renaming this to `$failoverConfigs`, for the same reason as above.

src/Client.php
quinthar: Per below, I think this is unnecessary and I recommend removing it.

src/Client.php
```diff
@@ -218,247 +325,236 @@ public function call($method, $headers = [], $body = '')
...
         // Make sure we succeed to create a socket
         if ($this->socket === false) {
             $socketError = socket_strerror(socket_last_error());
-            throw new ConnectionFailure("Could not connect to Bedrock primary or failover $socketError");
+            throw new ConnectionFailure("Could not connect to create socket: $socketError");
         }
 
-        // Configure this socket and connect to the primary host, failing over if necessary
-        socket_set_option($this->socket, SOL_SOCKET, SO_SNDTIMEO, ['sec' => $this->sendTimeout, 'usec' => 0]);
+        // Configure this socket and try all the possible hosts in order, till we find one that works or all of them fail
```
quinthar: I'm not sure what this comment means -- it's not doing anything to "try all possible hosts in order" -- this function just connects to a single host and throws an exception on failure.

src/Client.php
flodnv: +1, can we just use the value we use for auth?

src/Client.php
iwiznia: It contains the `timeout` which comes from apc. The ones in `hosts` are just host and port because they come from the configuration.

src/Client.php
```diff
@@ -15,37 +15,178 @@
  */
 class Client implements LoggerAwareInterface
 {
+    const HEADER_DELIMITER = "\r\n\r\n";
+    const HEADER_FIELD_SEPARATOR = "\r\n";
+
+    /**
+     * This is a list of hosts to use with it's ports and timeout. We store it here to not have to fetch it from APC on
+     * each request.
+     * @var array
```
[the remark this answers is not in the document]
iwiznia: It's not possible in PHP.

src/Client.php
iwiznia: In each web request we do tons of bedrock requests, removing this would mean reading and writing to the apcu cache in each bedrock (or auth in the future) request, this can be tons of times in one single request. It's also the same thing we do in auth right now.

src/Client.php
iwiznia: Exactly!

src/Client.php
iwiznia: Oh, right... I think convention is same line... will change.

src/Client.php
iwiznia: I can rename it to `mainHosts`, but this is a full configuration (which is just host and port really).

src/Client.php
iwiznia: We use different values for webrock and authrock (now), so I made this a config in case we want to keep that. We can use the same one in both if we want to, but given the option is already added, I'm leaving it.

src/Client.php
iwiznia: Given that's already added, I'm leaving it, doesn't hurt, you don't even need to specify it as it has a sane default (2 retries).

src/Client.php
iwiznia: I think you are right and I can pass it, will change.

src/Client.php
```diff
@@ -15,37 +15,178 @@
...
         // Store the configuration
-        self::$config = array_merge([
-            'host' => 'localhost',
-            'port' => 8888,
-            'failoverHost' => 'localhost',
-            'failoverPort' => 8888,
-            'connectionTimeout' => 300,
+        self::$defaultConfig = array_merge([
+            'clusterName' => 'bedrock',
+            'hosts' => ['localhost' => ['timeout' => 0, 'port' => 8888]],
```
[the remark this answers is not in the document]
iwiznia: I think this is a bad name, this is the timeout till when it's blacklisted... will change to `blackListedUntil`

src/Client.php
```diff
@@ -57,68 +198,21 @@ public static function getLogger()
      */
     public function setLogger(LoggerInterface $logger)
```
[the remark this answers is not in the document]
iwiznia: It's needed by the `LoggerAwareInterface` that this class implements.

src/Client.php
iwiznia: Again, because the `LogerAwareInterface` that this class implements is documented as so.

src/Client.php
iwiznia: No, `finally` runs even when no exception was thrown.

src/Client.php
iwiznia: lol, that's because you wrote it 😄

src/Client.php
iwiznia: I'll change the name, but you are misunderstanding things, `array_rand` returns a random key in an array, so we are not mixing 2 different things, it's all host names.

src/Client.php
iwiznia: No, see my comment above, will change naming though.

src/Client.php
iwiznia: hmmm, yeah BWM will not "learn" from cgi... not sure what we can do.. I could "re-read" from the cache after X time has passed. Does that sound good? If so, after how long?
Also, no, it's not a premature optimization, without the apc cache nothing would work as you would start each request with no knowledge. We save it so the information is saved for the next request.

iwiznia:
Updated

src/Client.php
```diff
@@ -15,37 +15,165 @@
...
+        $this->stats = $config['stats'];
+        $this->writeConsistency = $config['writeConsistency'];
+        $this->maxBlackListTimeout = $config['maxBlackListTimeout'];
+        $this->packetLength = $config['packetLength'];
+        $this->retries = $config['retries'];
+
+        // Make sure we have at least one host configured
+        $this->logger->debug('Bedrock\Client - Constructed', ['clusterName' => $this->clusterName, 'hosts' => $this->mainHostConfigs, 'failovers' => $this->failoverConfigs]);
+        if (empty($this->mainHostConfigs)) {
+            throw new BedrockError('Failed to construct Bedrock object');
```
flodnv: NAB message could be more descriptive

src/Client.php
```diff
@@ -168,43 +250,55 @@ public function call($method, $headers = [], $body = '')
...
+        $numTriesRemaining = $this->retries + 1;
         $response = null;
-        $lastTryException = null;
-        while($numTries-- && !$response) {
-            // Catch any connection failures and retry, but ignore non-connection failures.
+        $hosts = $this->getPossibleHosts();
+        $host = null;
+        while($numTriesRemaining-- && !$response && count($hosts)) {
+            reset($hosts);
+            $host = key($hosts);
```
flodnv: s/$host/$hostName for clarity?

src/Client.php
```diff
@@ -15,37 +15,165 @@
...
+
+    /**
+     *  @var null|string The last commit count of the node we talked to. This is used to ensure if we make a subsequent
+     * request to a different node in the same session, that the node waits until it is at least as "fresh" as the node
+     * we originally queried.
+     */
+    private $commitCount = null;
+
+    /**
+     *  @var null|resource Existing socket, if any.  It's reused each time, if possible.
```
flodnv: Is it really reused?

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
-    /**
-     * Hostname we attempt if the first didn't work.
-     *
-     *  @var null|string
-     */
-    private $failoverHost = null;
-
-    /**
-     * Port we attempt if the first didn't work.
+     * Sends the request on the existing socket, if possible, else it reconnects.
```
flodnv: `on the existing socket, if possible` is wrong, yeah?

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
-
-            return;
-        }
-        $this->getLogger()->debug('Opening new socket');
-
+        $this->logger->info('Bedrock\Client - Opening new socket', ['host' => $host]);
         // Try to connect to the requested host
-        $this->socketPID = posix_getpid();
+        if ($this->socket) {
+            socket_close($this->socket);
```
flodnv: why don't we close it after each request?

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
-        // Try up to 3 times, reconnecting each time
-        $attempts = 0;
-        while (++$attempts < 3) {
-            // Reconnect (if necessary) and send
-            $this->reconnect();
-            $bytesSent = socket_send($this->socket, $rawRequest, strlen($rawRequest), MSG_EOF);
-            if ($bytesSent === false) {
-                // Failed to send anything, let's try again (if we haven't exceeded our retry limit)
-                $errorCode = socket_last_error();
+        // Make sure bedrock is returning something https://github.com/Expensify/Expensify/issues/11010
```
flodnv: remove the link

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
+                throw new BedrockError("Error receiving data: $errorCode - $errorMsg");
+            }
+            if ($sizeDataOnSocket === 0 || strlen($dataOnSocket) === 0) {
+                throw new BedrockError('Bedrock response was empty');
+            }
+            $totalDataReceived += $sizeDataOnSocket;
+            $response .= $dataOnSocket;
+
+            // The first time are reading data from the socket, we need to extract the headers
+            // to ba able to get the size of the response
```
flodnv: typo `ba`

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
-        $response['code']     = intval($response['codeLine']);
-        $response['headers']  = [];
-        foreach ($responseHeaderLines as $responseHeaderLine) {
-            $nameValue = explode(':', $responseHeaderLine);
-            if (count($nameValue) != 2) {
-                $this->getLogger()->warning('Malformed response header, ignoring.', ['responseHeaderLine' => $responseHeaderLine]);
-            } else {
-                $response['headers'][ trim($nameValue[0]) ] = trim($nameValue[1]);
+            // We've seen occurrences of this happen when the string is not UTF-8. Forcing it fixes it.
+            // See https://github.com/Expensify/Expensify/issues/21805 for example.
```
flodnv: remove the link

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
+        ];
     }
 
     /**
-     * Receives a little bit more data.
+     * Parse a raw response from auth.
      *
-     * @return string The new data received.
+     * Note: $response is passed by reference so we don't store two copies in memory
+     * Note: this function can exit; and never return in case of token expiration
```
flodnv: what?

src/Client.php
```diff
@@ -218,247 +312,236 @@ public function call($method, $headers = [], $body = '')
...
-        while (strlen($rawResponseBody) < $contentLength) {
-            $rawResponseBody .= $this->recv();
-        }
-        if (strlen($rawResponseBody) != $contentLength) {
-            $this->getLogger()->warning('Server sent more content than expected, ignoring.', [
-                'Content-Length' => $contentLength,
-                'bodyLength' => strlen($rawResponseBody),
-                'response' => $rawResponseBody,
-            ]);
+    private function extractResponseHeaders($responseHeaderLines)
```
flodnv: `array $responseHeaderLines` ?

src/Client.php
iwiznia: We use it to ensure we destruct the socket, but the comment is wrong, updating.

src/Client.php
iwiznia: I thought about that, but something can fail in the middle and the user of the class could be catching the error and re-using the object to make a different request, so we need to make sure 100% there is not one open.

iwiznia: This way, I don't have to close the socket in each place we throw...

src/Client.php
iwiznia: At first I did, but then I re-added it, I know people don't have access to it, but is good context for us... we would loose that reference if I removed this (there's one more somewhere).

src/Client.php
iwiznia: Copy paste fail 💸

iwiznia:
Updated!

src/Client.php
quinthar: Well no, because we reuse the same socket, and only reconnect if the socket fails, correct?  So in theory, if your node doesn't die, you should load from the cache exactly once per PHP instance -- and only once per each time a node dies (which is rare).  So I still think this is overengineered.

src/Client.php
quinthar: Hm, well why not just overwrite the configuration options with the options from the cache?  This is confusing to have two nearly-identical copies of the same data.

src/Client.php
quinthar: Hm, let's just use a constant.  They were both arbitrarily picked, and clearly both work fine.  So let's just pick one and use a constant, thereby resisting adding extraneous complexity through useless knobs.

src/Client.php
quinthar: That's not a very strong defense.  If you wouldn't write it today from scratch, let's not include it just because it's there.

src/Client.php
quinthar: Ooh, duh.

src/Client.php
quinthar: lol, that's hilarious.

src/Client.php
quinthar: ahh, cool.  thanks for that clarification.

src/Client.php
quinthar: I would think we should:

1. Initialize from the config, so we have something for the "cold boot" scenario
2. Overwrite with whatever's in the cache, if anything is there, whenever establishing a connection

src/Client.php
```diff
@@ -218,247 +312,233 @@ public function call($method, $headers = [], $body = '')
...
-        $response['codeLine'] = array_shift($responseHeaderLines);
-        $response['code']     = intval($response['codeLine']);
-        $response['headers']  = [];
-        foreach ($responseHeaderLines as $responseHeaderLine) {
-            $nameValue = explode(':', $responseHeaderLine);
-            if (count($nameValue) != 2) {
-                $this->getLogger()->warning('Malformed response header, ignoring.', ['responseHeaderLine' => $responseHeaderLine]);
-            } else {
-                $response['headers'][ trim($nameValue[0]) ] = trim($nameValue[1]);
+        // Detect if we are using Gzip (TODO: can we remove this?)
```
quinthar: @tylerkaraszewski Does Bedrock support GZipped responses?  If not, I like the idea of removing this.

quinthar: On the other hand, I sorta thought GZipped responses were a really good thing that helped a log for large NVP objects.  Did we ever have this and did we remove it?

src/Client.php
```diff
@@ -15,37 +15,165 @@
...
 class Client implements LoggerAwareInterface
 {
+    const HEADER_DELIMITER = "\r\n\r\n";
+    const HEADER_FIELD_SEPARATOR = "\r\n";
+
+    /**
+     * @var array This is a list of hosts to use with it's ports and timeout. We store it here to not have to fetch it
+     * from APC on each request.
+     */
+    private static $cachedHosts = [];
```
quinthar: If you do keep this (which is really, really hope we don't) it should be named `$cachedHostConfigs` to match the other variables.

src/Client.php
```diff
@@ -475,4 +555,19 @@ private static function toUTF8($str)
...
+     * When a host fails, we blacklist that server for a certain amount of time, so we don't send requests to it when we
+     * know it's down. The blacklist time is a random amount of time between 1 second and the maxBlackListTimeout
+     * configuraion.
+     */
+    private function markHostAsFailed(string $host)
+    {
+        $blacklistedUntil = time() + rand(1, $this->maxBlackListTimeout);
+        self::$cachedHosts[$this->clusterName][$host]['blacklistedUntil'] = $blacklistedUntil;
+        if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING) && empty(self::$cachedHosts[$this->clusterName])) {
+            apcu_store('bedrockFailoverHosts-'.$this->clusterName, self::$cachedHosts[$this->clusterName]);
```
quinthar: I'd suggest putting `bedrockFailoverHosts-` into a constant, as it's used in multiple places.  Also, I'd suggest naming it to match exactly what it is storing, which at the moment is the (super terrible, please remove) `cachedHosts`.  But ideally it would just be named `bedrock-failoverHostConfigs` (and store `$failoverHostConfigs`).

src/Client.php
iwiznia: No, not correct. We close and reopen the socket each time. This is what auth was doing and reusing the socket is the (suspected) culprit of a bug we see in webrock (when a response from 2 or more request gets mixed up for some reason). 
Re-using the socket is a good optimization we can try out after this is done if we want, but introducing socket reuse to auth is too risky change IMO and we should test that separately.

src/Client.php
iwiznia: I agree, but auth doesn't do this now and the idea is to start using this lib for auth. We can reintroduce the socket reuse case later on in a more controlled fashion (ie: without all the other changes).

src/Client.php
iwiznia: I checked and we are not using this at all, in fact it doesn't even work. Given that, keeping this code is just bloat. If/when we want to really enable gzip, we can bring it back.

src/Client.php
```diff
@@ -218,247 +312,233 @@ public function call($method, $headers = [], $body = '')
...
+        if (is_null($json)) {
+            // This will remove unwanted characters.
+            // Check http://stackoverflow.com/a/20845642 and http://www.php.net/chr for details
+            for ($i = 0; $i <= 31; $i++) {
+                $body = str_replace(chr($i), '', $body);
+            }
+            $jsonStr = str_replace(chr(127), '', $body);
+
+            // We've seen occurrences of this happen when the string is not UTF-8. Forcing it fixes it.
+            // See https://github.com/Expensify/Expensify/issues/21805 for example.
```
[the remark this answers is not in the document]
iwiznia: Flo said the same thing, but I want to have this context somewhere, any idea where that might be? Being able to go back to that issue and see what is solving is useful information that I'd rather not lose.

src/Client.php
```diff
@@ -218,247 +312,233 @@ public function call($method, $headers = [], $body = '')
...
-        // Get more data from the socket
-        $buf = null;
-        $numRecv = @socket_recv($this->socket, $buf, 1024 * 1024, 0); // Read up to 1MB per call
-        if ($numRecv === false) {
-            $errorCode = socket_last_error();
-            $errorMsg = socket_strerror($errorCode);
-            throw new ConnectionFailure("Socket read failed: '$errorMsg' #$errorCode");
-        } elseif ($numRecv <= 0) {
-            throw new ConnectionFailure("Socket read failed: no data returned");
+        // Make sure bedrock is returning something https://github.com/Expensify/Expensify/issues/11010
```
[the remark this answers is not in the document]
iwiznia: See comment below of keeping it for context

iwiznia:
Updated.

src/Client.php
flodnv: Indeed, historically, reusing sockets has always been the source of bugs and has never worked fully (much less with auth). I'm on board with @iwiznia's plan of not reusing the socket here, and re-introducing that if we want later.

quinthar: Per [Slack convo](https://expensify.slack.com/archives/C03SSAQ7P/p1506360646000122) we're going to skip connection reuse for now.

quinthar:
Per [Slack convo](https://expensify.slack.com/archives/C03SSAQ7P/p1506360646000122), @iwiznia is going to remove `$cachedHosts` and just read the latest hostlist directly from the APC prior to every connection establishment.  Is this correct?

iwiznia:
Updated to remove the cache for the host configs. Please, re-review, but don't merge as I need to create the new version and re-test this.

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
-                // We sent something; can't retry else we might double-send the same request.  Let's make sure we sent the
-                // whole thing, else there's a problem.
-                if ($bytesSent == strlen($rawRequest)) {
-                    // Cool, we sent the whole thing.
-                    return;
-                } else {
-                    // Partial send -- abort
-                    throw new BedrockError('Sent partial request to Bedrock, aborting.');
-                }
+        // If cached hosts aren't set, get them from the APC cache. Then, we check the configuration there with the passed
```
quinthar: I think this comment is out of date?

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+        // configuration and if it's outdated (ie: it has different hosts from the one in the config), we reset it. This
+        // is so that we don't keep the old cache after changing the hosts or failover configuration.
+        if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING)) {
+            $apcuKey = self::APCU_CACHE_PREFIX.$this->clusterName;
+            $hostConfigs = apcu_fetch($apcuKey) ?: [];
+            $this->logger->info('Bedrock\Client - APC fetch failover hosts', $hostConfigs);
```
quinthar: In the service of consistency I'd vote:

1. Rename `bedrockHostStates-` to `bedrockHostConfigs-`
2. Change `APC fetch failover hosts` to `APC fetch host configs`

I know it seems picky, but there is just no reason not to just call the same thing the same name everywhere it appears.

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
-                } else {
-                    // Partial send -- abort
-                    throw new BedrockError('Sent partial request to Bedrock, aborting.');
-                }
+        // If cached hosts aren't set, get them from the APC cache. Then, we check the configuration there with the passed
+        // configuration and if it's outdated (ie: it has different hosts from the one in the config), we reset it. This
+        // is so that we don't keep the old cache after changing the hosts or failover configuration.
+        if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING)) {
+            $apcuKey = self::APCU_CACHE_PREFIX.$this->clusterName;
+            $hostConfigs = apcu_fetch($apcuKey) ?: [];
```
quinthar: Rename `$hostConfigs` to `$cachedHostConfigs`.  (Sorry, I'm having a really hard time figuring out what this logic does, because everything is name nearly identically, and the names keep changing.)

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+            $this->logger->info('Bedrock\Client - APC fetch failover hosts', $hostConfigs);
+
+            // If the hosts and ports in the cache don't match the ones in the config, reset the cache.
+            $savedHostsAndPort = [];
```
quinthar: With `$hostConfigs` renamed to `$cachedHostConfigs`, can you rename `$savedHostsAndPorts` to `$cachedHostsAndPorts`?  (Sorry, I'm just trying follow what this code does and am really having trouble.)

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+            foreach ($hostConfigs as $hostName => $config) {
+                $savedHostsAndPort[$hostName] = $config['port'];
+            }
+            asort($savedHostsAndPort);
+            $configHostsAndPort = [];
```
quinthar: Again, now it's using "config" in a new context that is totally different to how it was used previously.  Can you rename `$configHostsAndPort` to `$uncachedHostsAndPort` to more obviously contrast it to `$cachedHostsAndPorts` and avoid overloading the word "config" in a confusing way?

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+            foreach (array_merge($this->mainHostConfigs, $this->failoverHostConfigs) as $hostName => $config) {
+                $configHostsAndPort[$hostName] = $config['port'];
+            }
+            asort($configHostsAndPort);
+            if ($savedHostsAndPort !== $configHostsAndPort) {
+                $hostConfigs = array_merge($this->mainHostConfigs, $this->failoverHostConfigs);
+                $this->logger->info('Bedrock\Client - APC init failover hosts', $hostConfigs);
```
quinthar: Since they're called "host configs" (not "failover configs"), and since the command is "store" (not "init"), can you rename this comment from `Bedrock\Client - APC init failover hosts` to `Bedrock\Client - APC store host configs"?

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+        foreach ($hostNames as $hostName) {
+            $blackListedUntil = $hostConfigs[$hostName]['blacklistedUntil'] ?? null;
+            if (!$blackListedUntil || $blackListedUntil < time()) {
+                $nonBlackListedHosts[$hostName] = $hostConfigs[$hostName];
             }
         }
 
-        // Failed after three attempts
-        throw new BedrockError("Failed to send request to Bedrock after $attempts, aborting.");
+        $this->getLogger()->info('Bedrock\Client - Possible hosts', ['mainHostConfigs' => array_keys($nonBlackListedHosts)]);
```
quinthar: I think `mainHostConfigs` here should actually be `nonBlackListedHosts`?  Otherwise it seems like it's referring to a totally different variable.

src/Client.php
iwiznia: Removed the hosts, but leaving `init` because this is when it's initializing the cache due to a configuration change or because there was no data stored.

iwiznia:
Updated!

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+            }
+            asort($cachedHostsAndPorts);
+            $uncachedHostsAndPort = [];
+            foreach (array_merge($this->mainHostConfigs, $this->failoverHostConfigs) as $hostName => $config) {
+                $uncachedHostsAndPort[$hostName] = $config['port'];
+            }
+            asort($uncachedHostsAndPort);
+            if ($cachedHostsAndPorts !== $uncachedHostsAndPort) {
+                $cachedHostConfigs = array_merge($this->mainHostConfigs, $this->failoverHostConfigs);
+                $this->logger->info('Bedrock\Client - APC init host configs', $cachedHostConfigs);
```
quinthar: Rename `init` to `store` please

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+        $failoverHostNames = array_keys($this->failoverHostConfigs);
+        shuffle($failoverHostNames);
+        $mainHostName = array_rand($this->mainHostConfigs);
+        $hostNames = array_merge([$mainHostName], $failoverHostNames);
+
+        $nonBlackListedHosts = [];
+        foreach ($hostNames as $hostName) {
+            $blackListedUntil = $cachedHostConfigs[$hostName]['blacklistedUntil'] ?? null;
+            if (!$blackListedUntil || $blackListedUntil < time()) {
+                $nonBlackListedHosts[$hostName] = $cachedHostConfigs[$hostName];
```
quinthar: Do PHP associative arrays maintain order?  Assuming it's not blacklisted, we want to try `$mainHostName` first.  And we do in fact insert that into `$nonBlackListedHosts` first.  But is this guaranteed to be returned first from the PHP array?  Asked another way, if you do this:

```
$foo['a'] = 1;
$foo['b'] = 2;
reset($foo);
$first = key($foo);
```

Will `$first` *always* be `a`?

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
-                    // Partial send -- abort
-                    throw new BedrockError('Sent partial request to Bedrock, aborting.');
-                }
+        // We get the host configs from the APC cache. Then, we check the configuration there with the passed
+        // configuration and if it's outdated (ie: it has different hosts from the one in the config), we reset it. This
+        // is so that we don't keep the old cache after changing the hosts or failover configuration.
+        if ((!defined('TRAVIS_RUNNING') || !TRAVIS_RUNNING)) {
+            $apcuKey = self::APCU_CACHE_PREFIX.$this->clusterName;
+            $cachedHostConfigs = apcu_fetch($apcuKey) ?: [];
+            $this->logger->info('Bedrock\Client - APC fetch host configs', $cachedHostConfigs);
```
quinthar: This adds a large log line (containing the complete list of cached configs) to every single auth request -- is this necessary?  Can it be demoted to a `debug()`?

src/Client.php
```diff
@@ -218,247 +305,232 @@ public function call($method, $headers = [], $body = '')
...
+        foreach ($hostNames as $hostName) {
+            $blackListedUntil = $cachedHostConfigs[$hostName]['blacklistedUntil'] ?? null;
+            if (!$blackListedUntil || $blackListedUntil < time()) {
+                $nonBlackListedHosts[$hostName] = $cachedHostConfigs[$hostName];
             }
         }
 
-        // Failed after three attempts
-        throw new BedrockError("Failed to send request to Bedrock after $attempts, aborting.");
+        $this->getLogger()->info('Bedrock\Client - Possible hosts', ['nonBlacklistedHosts' => array_keys($nonBlackListedHosts)]);
```
quinthar: This adds a second large log line (containing the complete list of cached configs) to every single auth request -- is this necessary?  Can it be demoted to a `debug()`?

src/Client.php
iwiznia: Yes, you can read [here](http://php.net/manual/en/language.types.array.php) about it:
> An array in PHP is actually an ordered map
> Arrays are ordered. The order can be changed using various sorting functions.

src/Client.php
iwiznia: This just logs the server names, is not that long.

iwiznia:
Updated

src/Client.php
flodnv: Can we create a follow up issue to know what to do with this code? We keep discussing it over and over again... let's make it work, or kill it.

src/Client.php
```diff
@@ -15,37 +15,158 @@
...
+
+    /**
+     * The length of each packet read from the socket.
+     */
+    const PACKET_LENGTH = 16384;
+
+    /**
+     * The number of times to retry on a different server if the first one fails.
+     */
+    const RETRIES = 2;
```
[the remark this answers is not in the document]
iwiznia: You'd rather have a `2` in the middle of the code?

src/Client.php
iwiznia: Given that each server has a changing list of hosts that it uses, I think the log will help in some investigations. We can remove some logs later if we see they aren't, but I think we can benefit, especially especially now that we are rolling this out of this additional logging.

flodnv: sounds good

src/Client.php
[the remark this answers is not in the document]
flodnv: Well the 2 lines using this constant say:
```
        // We try +1 of the configured amount of retries, each time on a different valid server
        $numTriesRemaining = self::RETRIES + 1;
```

I'd change it to something like:
```
// Retry at most 3 times 
$numTriesRemaining = 3;
```
no?

iwiznia:
Updated.

PR #43, opened 2017-10-05
Remove unused timeout parameter

tylerkaraszewski:
@mea36 
cc @iwiznia 

Fixes: https://github.com/Expensify/Expensify/issues/63406

Removes unused timeout parameter. This is a problem as we now have a global `timeout` for all commands in bedrock. I'm aware there's an issue to bring back the previous version of `timeout`, but we'll need to do so with a different name.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -148,7 +148,7 @@ try {
             // Ready to get a new job
             try {
                 // Query the server for a job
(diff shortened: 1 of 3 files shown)


mea36:
Looks good, can you run `package.sh` to upgrade the package version? And then we need to upgrade the version of `Bedrock-PHP` in our web repo

tylerkaraszewski:
Related:
https://github.com/Expensify/Web-Secure/pull/610
https://github.com/Expensify/Web-Expensify/pull/18749

PR #46, opened 2017-10-11
Idempotent reads

iwiznia:
@flodnv, @deetergp  please review.

Needed for https://github.com/Expensify/Expensify/issues/63431

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -33,18 +33,60 @@ class DB extends Plugin
      * @param string $sql The query to run
      *
      * @return Response
+     * @deprecated Use read or write methods instead.
+     */
+    public function query($sql)
+    {
+        if (preg_match('/^\s*SELECT.*/i', $sql)) {
+            return $this->read($sql);
+        }
+        return $this->write($sql);
+    }
+
+    /**
+     * Executes a single read SQL query.
+     *
+     * @param string $sql The query to run
+     *
+     * @return Response
+     */
+    public function read($sql)
+    {
+        return $this->_runQuery($sql, true);
+    }
+
+    /**
+     * Executes a write read SQL query.
+     *
+     * @param string $sql The query to run
+     *
+     * @return Response
+     */
+    public function write($sql)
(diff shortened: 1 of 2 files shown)


src/DB.php
```diff
@@ -33,18 +33,60 @@ class DB extends Plugin
...
+     *
+     * @return Response
+     */
+    public function read($sql)
+    {
+        return $this->_runQuery($sql, true);
+    }
+
+    /**
+     * Executes a write read SQL query.
```
flodnv: "a write read SQL query" ? 😆

src/DB.php
```diff
@@ -33,18 +33,60 @@ class DB extends Plugin
...
+     *
+     * @param string $sql The query to run
+     *
+     * @return Response
+     */
+    public function write($sql)
```
flodnv: `string $sql`? `string $sqlQuery` and remove doc?

iwiznia:
Updated

src/DB.php
```diff
@@ -31,20 +31,52 @@ class DB extends Plugin
...
+        if (preg_match('/^\s*SELECT.*/i', $sql)) {
+            return $this->read($sql);
+        }
+        return $this->write($sql);
+    }
+
+    /**
+     * Executes a single read SQL query.
+     */
+    public function read(string $sql): Response
```
flodnv: so what happens if someone calls this with an INSERT or DELETE or UPDATE?

iwiznia: Nothing... and to clarify, it can make sense to say a write command is idempotent. An update might be idempotent (eg: `update something set value = 1`). 
Do you think we should throw or log or something else?

flodnv: Right.. it seems to me that maybe the function names aren't really reflecting what we want / what we're doing. Thoughts on renaming the methods like
- s/write/executeQuery
- s/read/executeIdempotentQuery

?

iwiznia:
Updated!

PR #51, opened 2017-10-23
Check for version file change in the safeToStartJob loop

mattabullock:
When BWM unfortunately gets stuck in the safeToStartJob (because of stuck jobs in the localJobsDB) the version watch file doesn't currently work, so I added a check in the loop to make sure it can still shut down properly when it's stuck in that loop.

Fixes https://github.com/Expensify/Expensify/issues/64494

# Tests
1. Run BWM with loadHandlerEnabled so it creates localJobsDB.sql
2. Fill the localJobsDB.sql with 10 dummy jobs with no closed date
3. Run BWM again with jobs queued, note that it keeps failing to start a job
4. `touch` the restart file, make sure it restarts correctly.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -399,10 +399,9 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
  *
  * Note: php's filemtime results are cached, so we need to clear
  *       that cache or we'll be getting a stale modified time.
- *
- * @return bool If version file changed
  */
- function checkVersionFile() {
+function checkVersionFile(string $versionWatchFile): bool
+{
      // Watch a version file that will cause us to automatically shut
      // down if it changes.  This enables triggering a restart if new
      // PHP is deployed.
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -117,6 +117,14 @@ try {
                 throw new Exception('are you in a chroot?  If so, please make sure /proc is mounted correctly');
             }
 
+            if (checkVersionFile()) {
+                $logger->info('Version watch file changed, stop processing new jobs');
+
+                // We break out of this loop and the outer one too. We don't want to process anything more,
+                // just wait for child processes to finish.
+                break 2;
+            }
+
             // Check if we can fork based on the load of our webservers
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -390,3 +390,31 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
...
+     // PHP is deployed.
+     //
+     // Note: php's filemtime results are cached, so we need to clear
+     //       that cache or we'll be getting a stale modified time.
+     clearstatcache(true, $versionWatchFile);
+     $newVersionWatchFileTimestamp = ($versionWatchFile && file_exists($versionWatchFile)) ? filemtime($versionWatchFile) : false;
+     $versionChanged = $versionWatchFile && $newVersionWatchFileTimestamp !== $versionWatchFileTimestamp;
+
+    if ($versionChanged) {
+        $logger->info('Version watch file changed, stop processing new jobs');
```
iwiznia: You added this log here, but also added it to the new caller of this method and left it in the previous place, you need to remove the other ones or this will get logged twice.

mattabullock:
Travis is still failing... what is `rake`?

cead22:
A ruby version of Make

mattabullock:
Any idea why it would fail in a travis build? I have no idea why it's used or anything.

cead22:
My guess is that Travis is struggling
image

iwiznia:
Try restarting the build... for some reason, I can't.

mattabullock:
@tgolen said we can just ignore it for now. It's failing to find the travis config because there isn't one yet.

That being said, let me finish testing it before merging (I'll do that this afternoon)

mattabullock:
Updated and tested fully!

bin/BedrockWorkerManager.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
iwiznia: Ballsy! You sure about this???

mattabullock: It helped me find a bug in my code, heh. 

All the params have default values so the only thing I can see that would make it fail is possibly Logger being null?

I tested it with load handler enabled and disabled and it worked fine both ways.

PR #52, opened 2017-10-31
Add a timeout exception

mea36:
This adds a timeout exception class.
Once you approve this or give a 👍  , I'll update the version number

Fixes https://github.com/Expensify/Expensify/issues/65018

Goes with https://github.com/Expensify/Web-Expensify/pull/19161/

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -2,10 +2,10 @@
 
 namespace Expensify\Bedrock;
 
+use Expensify\Bedrock\DB\Response;
 use Expensify\Bedrock\Exceptions\DB\FailedQuery;
 use Expensify\Bedrock\Exceptions\DB\UnknownError;
 use Expensify\Bedrock\Exceptions\Timeout;
-use Expensify\Bedrock\DB\Response;
 
 /**
  * Encapsulates the built-in DB plugin for Bedrock.
@@ -39,6 +39,7 @@ class DB extends Plugin
      * Executes a single SQL query.
      *
      * @param string $sql The query to run
+     *
      * @deprecated Use run method instead.
      */
     public function query($sql): Response
@@ -52,8 +53,8 @@ class DB extends Plugin
     /**
      * Executes an SQL query.
      *
-     * @param string $sql The query to run
-     * @param bool $idempotent Is this command idempotent? If the command is run twice is the final result the same?
+     * @param string $sql        The query to run
+     * @param bool   $idempotent Is this command idempotent? If the command is run twice is the final result the same?
      *
      * @throws FailedQuery
      * @throws UnknownError
--- a/src/DB.php
+++ b/src/DB.php
@@ -4,6 +4,7 @@ namespace Expensify\Bedrock;
 
 use Expensify\Bedrock\Exceptions\DB\FailedQuery;
 use Expensify\Bedrock\Exceptions\DB\UnknownError;
+use Expensify\Bedrock\Exceptions\Timeout;
 use Expensify\Bedrock\DB\Response;
 
 /**
@@ -26,6 +27,13 @@ class DB extends Plugin
      * @var int
      */
     const CODE_QUERY_FAILED = 502;
+
+    /**
+     * Query timed out
+     *
+     * @var int
+     */
+    const CODE_QUERY_TIMEOUT = 555;
 
     /**
      * Executes a single SQL query.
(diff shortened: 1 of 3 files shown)


iwiznia:
So, this will be used by `lib/BedrockPlugins/Concierge.php`? Will it ever be caught and handled specifically? Also, `lib/BedrockPlugins/Concierge.php` should not be in web-e, should be in this repo.

mea36:
No, this will be used by anything in Web-E that thinks there might be a timeout.  The uses I've seen for this so far are DB::query and Concierge.



>  Also, lib/BedrockPlugins/Concierge.php should not be in web-e, should be in this repo.

I disagree, the Concierge plugin is not open source and therefore shouldn't be in this repo

iwiznia:
> I disagree, the Concierge plugin is not open source and therefore shouldn't be in this repo

Goddamnit, you are right...

> this will be used by anything in Web-E that thinks there might be a timeout. The usages I've seen for this so far are DB::query and Concierge

Can you create those PRs too, is hard to review this without that context.

src/Exceptions/Timeout.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
```
iwiznia: Declare strict types.

mea36:
Updated

iwiznia:
All looks good, but you need to create a new version.

mea36:
@iwiznia I created the new version

mea36:
I updated the version number

src/DB.php
```diff
@@ -66,6 +75,10 @@ public function run(string $sql, bool $idempotent): Response
             throw new FailedQuery("Query failed: {$response->getError()}");
         }
 
+        if ($response->getCode() === self::CODE_QUERY_TIMEOUT) {
+            throw new Timeout("Query timeoout: {$response->getError()}");
```
quinthar: Typo: `timeoout`

quinthar:
FYI I've asked that we hold on this one.  I'm concerned this is an extraneous exception that should be handled like the other `BedrockErrors` -- in the Concierge case it means "just ignore this call to `GetResponse` because it's screwed up somehow, but it doesn't matter because it's optional".

quinthar:
FYI, pulling in the discussion here: https://github.com/Expensify/Bedrock-PHP/pull/52#issuecomment-344134302

quinthar:
Oh crap now I see it's already been deployed.  Bah.  You will all think I'm crazy, but I think this is the wrong direction to take the code and would love to revert.  Extraneous complexity like this is so amazingly cheap to add, and so expensive to remove.  Let's resist this temptation at every turn and set a high bar to justifying each new exception, as most of them are never used, and those that are -- are frequently used wrong (as I think this is the case here).

fnwbr:
@quinthar 

I think you wanted to link to something else in https://github.com/Expensify/Bedrock-PHP/pull/52#issuecomment-344134441... 

Regarding this topic: I don't think it's good practice nor a good argument to say that we don't want this exception just because in that very one use case of Concierge we would handle the `Timeout` exception in the same way as any other type of `BedrockError`. Bedrock and Bedrock-PHP are supposed to be usable by anyone out there. Everything that we're doing with Concierge is our very own implementation of those.

Or let me ask this in a different way: Why do we even even add a separate response code for timeouts in the C++ layer then, when we basically want to forbid treating those differently?
I think the PHP implementation should represent the actual code of Bedrock as closely as possible, and that means a separate type of exception for every separate response code.

Since all those exceptions extend from the same `BedrockError` base class we also do **not** take away the possibility of treating them all the same way. So if (we for example) actively decide to not make a difference between timeout and other failures we can just keep looking out for `BedrockError`.

Can you clarify what exactly you are scared about in terms of this being "so expensive to remove"? Are you worried about the (theoretical) case of us removing the timeout feature and therefore having to go through the PHP codebase, removing all the `catch (TimeoutException $e)` instances?

iwiznia:
I think I agree with @quinthar, pending on confirmation of [this comment](https://github.com/Expensify/Expensify/issues/66118#issuecomment-344305358). If when the 555 error code is thrown in auth we are sure the command will not be processed, then I agree we should remove, I initially thought that wasn't the case, but now I'm not entirely sure.

fnwbr:
Following up on a 1:1 with @quinthar: Let's apply a general policy of never creating exceptions that aren't caught + actively refactoring away any exception that isn't caught here.
We should also discourage creating extraneous exceptions, as in many cases they wrongly encourage overly precise error handling at the expense of more general error handling.

In the end we want to:
- Revert this PR
- Also remove the `FailedQuery` and `UnknownError` exceptions (only **1** usage in Web-Expensify that I could find)

fnwbr:
I'm going to re-open the original issue https://github.com/Expensify/Expensify/issues/65018 for that; but I can't work on this right now.

@mea36 @iwiznia @pecanoro If any of you guys wants to go ahead and create the revert please feel free to; if you want me to do this please give me a few days. If you really need the changes I just listed above right away please let me know. For now I'm going ahead and assigning the issue to myself.

quinthar:
Thanks @fnwbr!

mea36:
I wrote this, I'll handle the revert.  Thanks for the discussion guys

PR #53, opened 2017-11-02
Add log line for when running jobs over target

mattabullock:
We want a better log line for when there are more running jobs than max jobs so we can accurately track when we're being throttled by AIMD.

# Test
1. Create 30 jobs
2. Set ended to null in the last 10
3. Run BWM with `--enableLoadHandler`
4. Look for this line in the logs:
```
Nov  2 21:14:52 vagrant-ubuntu-trusty-64 php: s7W5Q4 vendor/bin/BedrockWorkerManager.php we@dont.know !script! ?api? [info] Not queueing job, number of running jobs is above the target ~~ numActive: '10' target: '10'
```

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -308,7 +308,6 @@ try {
             $logger->warning("Failed to get job");
         }
     }
-
 } catch (Exception $e) {
     $message = $e->getMessage();
     $logger->alert('BedrockWorkerManager.php exited abnormally', ['exception' => $e]);
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -386,6 +386,8 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
         $logger->info("Jobs are slowing down, halving the target", ['newBatchAverageTime' => $newBatchAverageTime, 'oldBatchAverageTime' => $oldBatchAverageTime, 'numActive' => $numActive, 'target' => $target]);
     }
 
+    $logger->info("Not queueing job, number of running jobs is above the target", ['numActive' => $numActive, 'target' => $target]);
+
     // Don't authorize BWM to call GetJobs
     return [false, $target];
 }
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -238,6 +238,7 @@ try {
                     $stats->benchmark('bedrockJob.finish.'.$job['name'], function () use ($workerName, $bedrock, $jobs, $job, $extraParams, $logger, $localDB, $enableLoadHandler) {
                         $worker = new $workerName($bedrock, $job);
                         $childPID = getmypid();
+                        $localJobID = 0;
 
                         // Open the DB connection after the fork in the child process.
                         if ($enableLoadHandler) {
@@ -304,7 +305,7 @@ try {
         } elseif ($response['code'] == 303) {
             $logger->info("No job found, retrying.");
         } else {
-            $logger->warning("Failed to get job", ['codeLine' => $job['codeLine']]);
+            $logger->warning("Failed to get job");
         }
     }
 
@@ -333,11 +334,12 @@ $logger->info('Stopped BedrockWorkerManager');
  *
  * @return array First value is whether or not it is safe to start a new job, second is an updated $target value.
  */
-function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $enableLoadHandler, bool $debugThrottle, Psr\Log\LoggerInterface $logger) : array
+function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $enableLoadHandler, bool $debugThrottle, Psr\Log\LoggerInterface $logger): array
 {
     // Allow for disabling of the load handler.
     if (!$enableLoadHandler) {
         $logger->info("Load handler not enabled");
+
         return [true, $target];
     }
 
@@ -346,6 +348,7 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
     if ($numActive < $target) {
         // Still in a safe zone, don't worry about load
         $logger->info("Safe to start new job", ["numActive" => $numActive, "target" => $target]);
+
         return [true, $target];
     }
 
@@ -355,6 +358,7 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
         // Wait until we finish at least two batches of our target so we can evaluate its speed,
         // before expanding the batch.
         $logger->info("Haven't finished two batches of target, not queuing job", ["numActive" => $numActive, "target" => $target]);
+
         return [false, $target];
     }
 
@@ -380,7 +384,7 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
 
         // Authorize one more job given that we've just increased the target by one.
         return [true, $target];
-    } else if ($newBatchAverageTime > $maxSafeTime || $newBatchAverageTime < 1.5 * $oldBatchAverageTime) {
+    } elseif ($newBatchAverageTime > $maxSafeTime || $newBatchAverageTime < 1.5 * $oldBatchAverageTime) {
         // Things seem to be slowing down, pull our target back a lot
         $target = max(floor($target / 2), $minSafeJobs);
         $logger->info("Jobs are slowing down, halving the target", ['newBatchAverageTime' => $newBatchAverageTime, 'oldBatchAverageTime' => $oldBatchAverageTime, 'numActive' => $numActive, 'target' => $target]);
@@ -393,7 +397,6 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
 }
 
 /**
- *
  * Watch a version file that will cause us to automatically shut
  * down if it changes.  This enables triggering a restart if new
  * PHP is deployed.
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -333,11 +333,12 @@
  *
  * @return array First value is whether or not it is safe to start a new job, second is an updated $target value.
  */
-function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $enableLoadHandler, bool $debugThrottle, Psr\Log\LoggerInterface $logger) : array
+function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int $minSafeJobs, bool $enableLoadHandler, bool $debugThrottle, Psr\Log\LoggerInterface $logger): array
 {
     // Allow for disabling of the load handler.
     if (!$enableLoadHandler) {
         $logger->info("Load handler not enabled");
+
```
quinthar: Why this extra line?  Can you remove this?  https://github.com/Expensify/Style-Guide/blob/master/php.md#blank-lines

bin/BedrockWorkerManager.php
```diff
@@ -346,6 +347,7 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
     if ($numActive < $target) {
         // Still in a safe zone, don't worry about load
         $logger->info("Safe to start new job", ["numActive" => $numActive, "target" => $target]);
+
```
quinthar: Same: https://github.com/Expensify/Style-Guide/blob/master/php.md#blank-lines

bin/BedrockWorkerManager.php
```diff
@@ -355,6 +357,7 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
         // Wait until we finish at least two batches of our target so we can evaluate its speed,
         // before expanding the batch.
         $logger->info("Haven't finished two batches of target, not queuing job", ["numActive" => $numActive, "target" => $target]);
+
```
quinthar: Same: https://github.com/Expensify/Style-Guide/blob/master/php.md#blank-lines

bin/BedrockWorkerManager.php
```diff
@@ -380,18 +383,19 @@ function safeToStartANewJob(LocalDB $localDB, int $target, int $maxSafeTime, int
 
         // Authorize one more job given that we've just increased the target by one.
         return [true, $target];
-    } else if ($newBatchAverageTime > $maxSafeTime || $newBatchAverageTime < 1.5 * $oldBatchAverageTime) {
+    } elseif ($newBatchAverageTime > $maxSafeTime || $newBatchAverageTime < 1.5 * $oldBatchAverageTime) {
         // Things seem to be slowing down, pull our target back a lot
         $target = max(floor($target / 2), $minSafeJobs);
         $logger->info("Jobs are slowing down, halving the target", ['newBatchAverageTime' => $newBatchAverageTime, 'oldBatchAverageTime' => $oldBatchAverageTime, 'numActive' => $numActive, 'target' => $target]);
     }
```
quinthar: Same: https://github.com/Expensify/Style-Guide/blob/master/php.md#blank-lines  - perhaps put the info under the `// Don't authorize` comment beneath it?

bin/BedrockWorkerManager.php
mattabullock: It goes against the style checker that was recently implemented on this repo because we need them before return statements. @tgolen could weigh in on this one more.

quinthar: Can we fix the style checker to not encourage extraneous lines that are against the style guide?

iwiznia: Maybe we can (not sure what rule is applying this or if it can be configured), but should we? The style checker requires a blank line before a `return` when the previous line doesn't end with a `{`. All our codebase is already like that, so changing it now will make all files to be reformatted...

quinthar: To clarify, you are saying we've configured our style checker to enforce something contrary to our style guide, and then updated the codebase with an invalid style?  In general, I think that sounds like a mistake and a big waste of time.  At a minimum, can we agree not to do that again?

Regardless, I think the guide is right, and we should enforce the actual agreed upon guide.  If that means undoing a big mistake, well, that sucks, but that's why we try to avoid big mistakes.

iwiznia: No, I'm not saying that. The style checker was in place, then we updated our style guide to go against what the style checker did.
I'm not passionate about these blank lines, so I don't care if we update the checker or the style guide.

quinthar: ooh, got it.  Thanks.  I'd prefer to update the checker, as I think the guide is right.

quinthar:
Alright, continuing the saga of how to `Make AIMD Great!`, I don't know if this log line is enough.  If the problem is the AIMD algorithm just isn't authorizing jobs as fast as we expect, there can only be two explanations:

1. The algorithm is generating the wrong output
2. We have mistaken assumptions about the inputs

To determine which of those is the case, I think we need to log all the inputs/outputs of the algorithm and then go through it by hand to make sure it's correctly calculating everything -- and that the inputs are what we'd expect.  Specifically, I suggest we log the full state of the AIMD algorithm every time we call `safeToStartANewJob`, including: 

* The current number of running jobs
* The current target
* The average job speed of the last 2 batches

The goal is to confirm that we keep adding jobs so long as doing so doesn't slow down the processing speed.  Once adding jobs makes them go slower, then we halve the target and start gradually increasing again.

If we do this and see that for some reason we aren't adding jobs despite the job speeds remaining constant, then we've found a bug to fix.

Regardless, we need to pull enough data together into a single place to analyze it and come up with theories of what the bug is and how to fix it.

iwiznia:
Didn't we find that stuck jobs (which are still a thing) were skewing the results? Have we done anything to fix that?

mattabullock:
>To determine which of those is the case, I think we need to log all the inputs/outputs of the algorithm and then go through it by hand to make sure it's correctly calculating everything -- and that the inputs are what we'd expect. Specifically, I suggest we log the full state of the AIMD algorithm every time we call safeToStartANewJob, including: [...]

I've done that mostly already by tailing the logs and making graphs in excel, but I think since we were missing the log line in this PR it's tough to get a good read when it hits the max and stays there for a bit. I will also make sure I start grabbing the times from the log lines.

If we can get this change deployed today (we can just get the composer change CP'd to prod to speed this up). I can pull in the logs tonight and get a first look at the graph to see if we can find anything simple out from it.

If there is a bug it's an easy step forward, but I think

If I had to guess, I'd say that it's probably the times being too variable and cutting down too quickly and if that's the case I think we're going to need to come up with a better way to learn if we need to cut back on the target. However I'll hold judgement til we get numbers.

>Didn't we find that stuck jobs (which are still a thing) were skewing the results? Have we done anything to fix that?

They're skewing results but not enough to cause the dip in performance we've seen.

cead22:
To clarify, this PR doesn't output every data point available to the algorithm in a single log line right? Why, Won't that make evaluating the results harder?

> If we can get this change deployed today

Can you lead the charge on making this happen? I don't think there's any reason for us not to CP the composer change to production and start testing this as soon as possible.

> I've done that mostly already by tailing the logs and making graphs in excel, but I think since we were missing the log line in this PR it's tough to get a good read when it hits the max and stays there for a bit. I can pull in the logs tonight and get a first look at the graph to see if we can find anything simple out from it.
>If there is a bug it's an easy step forward, but I think

What do you expect to see in the new charts? what are you charting? how do you expect bugs to look like here, or are you hoping that the visualization will highlight something obvious?

> If I had to guess, I'd say that it's probably the times being too variable and cutting down too quickly

Is there a way for us to try to prove or disprove this theory by adding some extra logging in this PR or how do you expect to determine if the times are too variable?

mattabullock:
>To clarify, this PR doesn't output every data point available to the algorithm in a single log line right? Why, Won't that make evaluating the results harder?

I tried to make it only output relevant information, but going on what I said we should add the average times to this as well.

>Can you lead the charge on making this happen? I don't think there's any reason for us not to CP the composer change to production and start testing this as soon as possible.

Yes we should just CP this, we need to get past the empty lines thing but I really don't care either way, we just need to come to a conclusion.

>What do you expect to see in the new charts? what are you charting? how do you expect bugs to look like here, or are you hoping that the visualization will highlight something obvious?

I've been charting number of jobs currently running and target as a function of time. This is helpful to give us a view of how many jobs are running every time the algorithm is running and what the target they should be at is. 

The things we can verify using these graphs:

AIMD is too restrictive: Compare the target times to the bedrock jobs run graph to see if we're raising the target fast enough without it cutting down too quickly. This is pretty general but it counters the one below it at a high level. My theory is in the third 

BWM is not queuing jobs fast enough: Compare the active jobs to the target during the times of the peaks in the bedrock jobs run graph. If the active jobs aren't constantly at the target, we're not queueing jobs fast enough. We've kind of beat this one into the ground, but I think we'll have a better picture of it with this last log line, since it accounts for all the spots where we're at or above the target and won't be queueing anything else.

Times are too sporadic - This is my theory. With the huge difference in job speed because of variation in jobs we're cutting back too frequently because times aren't predictable enough. We'll be able to tell if the average time isn't relatively constant. Not having a good idea I can't give a good example of what this will look like, except that after adding it to this log line we'll see it jumping around during low load times by an amount over 50% (the amount of time that we will increase if it's over and same with in high load times. One issue with adding it to _all_ the log lines is that if we do that, we'll have to calculate the time every time we run the algorithm which slows it down a little bit. To get a good look at this we can graph the average batch time against the time when the algorithm was run. If we want to see the variability in the times, we can take the difference between each two points in the graph and graph those over time and see how high the variability is in percent or in raw ms.

JLiebsExpensify:
Awesome, thanks for holding laying out the theories!

mattabullock:
Fixed the travis tests. If we're fine with the blank space then if this could get approved/merged asap that would be appreciated!

cead22:
Given that the style checked will fail if we remove the blank lines know, I suggest creating an issue to fix that separate from this, merging this and CPing it so that we can begin using the log line to identify any issuea asap

cead22:
Here's the issue for the style checker https://github.com/Expensify/Expensify/issues/65684

PR #55, opened 2017-11-08
Add connection retries as configuration option to bedrock client

brandonhub:
@iwiznia , will you please review this?

This PR adds the number of retries the bedrock client is allotted to connect to the server as a configuration options.

Fixes https://github.com/Expensify/Expensify/issues/64873 and https://github.com/Expensify/Expensify/issues/65058

Note, the remaining phan errors are a bug with phan. I believe they can be ignored.
Also note, that the there is an accompanying [web pr](https://github.com/Expensify/Web-Expensify/pull/19151).

# Tests
None

# QA
None

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -30,20 +30,20 @@ class Client implements LoggerAwareInterface
 
     /**
      * @var array This is a default configuration applied to all instances of this class. They can be overriden in the
-     * constructor.
+     *            constructor.
      */
     private static $defaultConfig = [];
 
     /**
      *  @var string[] The last commit count of the node we talked to, keyed by the cluster name. This is used to ensure
-     * if we make a subsequent request to a different node in the same session, that the node waits until it is at least
-     * up to date with the commits as the node we originally queried.
+     *                if we make a subsequent request to a different node in the same session, that the node waits until it is at least
+     *                up to date with the commits as the node we originally queried.
      */
     private static $commitCount = [];
 
     /**
      * @var null|string Name of the bedrock cluster we are talking to. If you have more than one bedrock cluster, you
-     * can pass in different names for them in order to have separate statistics collected and caches of failed servers.
+     *                  can pass in different names for them in order to have separate statistics collected and caches of failed servers.
      */
     private $clusterName = null;
 
@@ -59,7 +59,7 @@ class Client implements LoggerAwareInterface
 
     /**
      *  @var array List of failovers we attempt if the first didn't work. We randomize the list and try on several of
-     * them (depending on the number of retries configured).
+     *             them (depending on the number of retries configured).
      */
     private $failoverHostConfigs = [];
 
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -497,13 +502,14 @@ private function receiveResponse()
...
-     * @return array the decoded json, or null on error
+     * @return array|null the decoded json, or null on error
+     *
      * @throws BedrockError
      */
     private function parseRawBody(array $headers, string $body)
     {
         // Detect if we are using Gzip (TODO: can we remove this?)
-        if (isset($responseHeaders['Content-Encoding']) && $headers['Content-Encoding'] === 'gzip') {
+        if (isset($headers['Content-Encoding']) && $headers['Content-Encoding'] === 'gzip') {
```
iwiznia: lol, thanks phan!

src/Client.php
```diff
@@ -74,6 +74,11 @@ class Client implements LoggerAwareInterface
     private $readTimeout;
 
     /**
+     * @var int The number of times to retry on a different server if the first one fails.
```
iwiznia: So, we will actually try the amount of servers configured here, so the comment is slightly wrong and also maybe the name of the variable/config. I'd say either call this `totalTries` or add 1 to the configured number [here](https://github.com/Expensify/Bedrock-PHP/pull/55#diff-ff73e042e738204c6da009e2ed19f783R267). Thoughts?

brandonhub: Ah I see, I think plus 1 is a good option, I think that's how it was done before

quinthar:
To clarify, why do we want to configure retry counts manually; why not just base the number of retries on how many failover nodes there are?

brandonhub:
I was under the impression that the number of retires wasn't necessarily directly related to the number of failover nodes configured. [this comment](https://github.com/Expensify/Bedrock-PHP/blob/master/src/Client.php#L61-L62) seems to suggest that failover nodes and retries are separate configurations by design. We can change that of course. @iwiznia any thoughts?

src/Client.php
```diff
@@ -159,6 +164,7 @@ public static function configure(array $config)
             'stats' => new NullStats(),
             'writeConsistency' => 'ASYNC',
             'maxBlackListTimeout' => 1,
+            'retries' => 4,
```
righdforsa: Since we know that a deploy will close the ports of half the cluster, it seems reasonable to set the max retries to ceiling(half the cluster), which would be 6 for webrock, or 3 for auth bedrock. Is it feasible to determine which cluster we're connecting to at this point?

quinthar:
Ya, I hate extraneous knobs -- let's just make it figure out the right answer automatically and do that.

iwiznia:
> why not just base the number of retries on how many failover nodes there are?

Oh great idea! We could get the total number of configured failovers (not the ones that are "alive" right now, but the total number in the original configuration) and base it on that. I like that much more than passing it down, it also would automatically adjust if we add/remove servers.

brandonhub:
> We could get the total number of configured failovers

Seems like a good idea. When you say configured failovers are you referring to what is passed into [this parameter](https://github.com/Expensify/Bedrock-PHP/blob/master/src/Client.php#L123)? I guess that doesn't seem right because that parameter seems to always be set to an array with only a single configuration inside. If that parameter isn't a valid representation of the number of failover nodes configured, where else could I get that number?

quinthar:
What about just retrying as many times as there are configured nodes to connect to?  So look at the failover configuration and see how big the cluster is, just try that many times.

iwiznia:
> Seems like a good idea. When you say configured failovers are you referring to what is passed into this parameter?

No, I was talking about [this param](https://github.com/Expensify/Bedrock-PHP/blob/master/src/Client.php#L123).

> What about just retrying as many times as there are configured nodes to connect to? So look at the failover configuration and see how big the cluster is, just try that many times.

So, try in all servers? We can do that too...

JLiebsExpensify:
Ya all servers. I'm not sure why we thought to do only half, and by doing
only half there is a rare edge case of finding no server at the exact
moment we do a master flip.

brandonhub:
Updated, I'm aware the packaging is wrong, I'll fix it once the change gets approved

composer.json.orig
```diff
@@ -0,0 +1,35 @@
+{
```
iwiznia: Delete this file, I have no idea what it is...

src/Client.php
```diff
@@ -259,13 +256,12 @@ public function call($method, $headers = [], $body = '')
...
 
-        // We try 3 times, each time on a different valid server
-        $numTriesRemaining = 3;
         $response = null;
         $hostConfigs = $this->getPossibleHosts();
         $hostName = null;
-        while($numTriesRemaining-- && !$response && count($hostConfigs)) {
+        while (!$response && count($hostConfigs)) {
             reset($hostConfigs);
+            $numRetriesLeft = count($hostConfigs) - 1;
```
iwiznia: Do we even need the numRetries? Given that there's already [this](https://github.com/Expensify/Bedrock-PHP/pull/55/files#diff-ff73e042e738204c6da009e2ed19f783R297)...

brandonhub: So I guess we could get rid of it but we would have to change some other things too. For starters we use `numRetriesLeft` to determine what sort of log to print, which isn't a big deal, we could just change the log lines in [this if else statement](https://github.com/Expensify/Bedrock-PHP/pull/55/files#diff-ff73e042e738204c6da009e2ed19f783R277) and [this if else statement](https://github.com/Expensify/Bedrock-PHP/pull/55/files#diff-ff73e042e738204c6da009e2ed19f783R285) and combine them into two generic logs I suppose.  And I'm guessing you were thinking [the line you linked](https://github.com/Expensify/Bedrock-PHP/pull/55/files#diff-ff73e042e738204c6da009e2ed19f783R297) would serve as the notification that we ran out of retry attempts? But then we wouldn't differentiate between a `BedrockError` and a `ConnectionError`, we just throw a connection error despite what error actually caused `$response` to be null. So I guess I'm a little unclear on what alterative you had in mind.

PR #56, opened 2017-11-10
Ensure safeToStartANewJob returns an int as the target

flodnv:
@mattabullock 
cc @iwiznia 
See https://github.com/Expensify/Expensify/issues/65705 -- logs are throwing `Argument 2 passed to safeToStartANewJob() must be of the type integer, float given, called in /git/expensify.com/vendor/expensify/Bedrock-PHP/bin/BedrockWorkerManager.php on line 129'`. This is the only place I could find that doesn't return an int (funnily enough, `floor` does not return an `int`) 

Not tested. LMK if this is right, and we can build the package.


iwiznia:
You need to create a new version... but anyway waiting for @mattabullock

PR #61, opened 2017-11-20
Reuse sockets

iwiznia:
Fixes https://github.com/Expensify/Expensify/issues/66144

This introduces socket reutilization to the bedrock Client class. In order to support being used in a multi threaded environment, like in BWM, it stores the sockets per pid, cluster and host name.

/cc @quinthar

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -289,7 +289,7 @@ class Client implements LoggerAwareInterface
             reset($hostConfigs);
             $numRetriesLeft = count($hostConfigs) - 1;
 
-            // If we already have a socket for this pid and cluster, then we first try to reuse it
+            // If we already have a socket for this instance, then we first try to reuse it
             if ($this->socket) {
                 $hostName = $this->lastHost;
             } else {
--- a/src/Client.php
+++ b/src/Client.php
@@ -35,16 +35,21 @@ class Client implements LoggerAwareInterface
     private static $defaultConfig = [];
 
     /**
-     *  @var string[] The last commit count of the node we talked to, keyed by the cluster name. This is used to ensure
-     *                if we make a subsequent request to a different node in the same session, that the node waits until it is at least
-     *                up to date with the commits as the node we originally queried.
+     * @var Client[] An array of all instances of this class that were created, one per different config passed.
      */
-    private static $commitCount = [];
+    private static $instances = [];
 
     /**
-     *  @var array Existing sockets. This array is keyed by cluster name, inside it has an array with hostName and socket.
+     *  @var string The last commit count of the node we talked to. This is used to ensure if we make a subsequent
+     *              request to a different node in the same session, that the node waits until it is at least
+     *              up to date with the commits as the node we originally queried.
      */
-    private static $sockets = [];
+    private $commitCount = '';
+
+    /**
+     *  @var resource|null Socket to the server.
+     */
+    private $socket = null;
 
     /**
      * @var null|string Name of the bedrock cluster we are talking to. If you have more than one bedrock cluster, you
@@ -115,7 +120,7 @@ class Client implements LoggerAwareInterface
      *
      * @throws BedrockError
      */
-    public function __construct(array $config = [])
+    private function __construct(array $config = [])
     {
         $config = array_merge(self::$defaultConfig, $config);
         $this->clusterName = $config['clusterName'];
@@ -136,12 +141,30 @@ class Client implements LoggerAwareInterface
     }
 
     /**
+     * Returns an instance of this class for the specified configuration. It will return the same instance if the same
+     * configuration was passed, unless clearInstancesAfterFork is called.
+     */
+    public static function get(array $config = []): Client
+    {
+        $config = array_merge(self::$defaultConfig, $config);
+        ksort($config);
+        $hash = sha1(print_r($config, true));
+        if (isset(self::$instances[$hash])) {
+            return self::$instances[$hash];
+        }
+        $instance = new Client($config);
+        self::$instances[$hash] = $instance;
+
+        return $instance;
+    }
+
+    /**
      * After forking, you need to call this method to make sure the forks don't share the same socket and instead open
      * a new connection.
      */
-    public static function clearSocketsAfterFork()
+    public static function clearInstancesAfterFork()
     {
-        self::$sockets = [];
+        self::$instances = [];
     }
 
     /**
@@ -221,8 +244,8 @@ class Client implements LoggerAwareInterface
         $timeStart = microtime(true);
 
         // Include the last CommitCount, if we have one
-        if (isset(self::$commitCount[$this->clusterName])) {
-            $headers['commitCount']  = self::$commitCount[$this->clusterName];
+        if ($this->commitCount) {
+            $headers['commitCount']  = $this->commitCount;
         }
 
         // Include the requestID for logging purposes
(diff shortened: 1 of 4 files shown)


iwiznia:
I'll create the new version when it's reviewed.

src/Client.php
```diff
@@ -135,10 +135,22 @@ public function __construct(array $config = [])
         }
     }
 
-    public function __destruct()
+    /**
+     * Closes all sockets for all clusters for the given pid.
```
flodnv: NAB: `... for the given process id`

iwiznia:
> , Phan is failing 😬

Bad phan error, we check if the constant is defined _right there_.

flodnv:
Ah right, seems to be the same bug as the nullness check (when we check if param is not null to pass to a function, yet Phan says we pass null, which is logically impossible)

quinthar:
Hm, this is a much more extensive solution than I expected.  I was thinking we would just:

1. change [this line](https://github.com/Expensify/Bedrock-PHP/blob/master/src/Client.php#L331) to avoid re-connecting if we already have a socket (I think it was originally implemented this way actually)
2. Update [here](https://github.com/Expensify/Bedrock-PHP/blob/master/bin/BedrockWorkerManager.php#L216) to close the socket when forking?

It's not *exactly* a 2 line change, but it's pretty close.  Thoughts on this dramatically simpler solution?

quinthar:
One possible explanation I can think of for this different solution is that it allows for reuse between different `Bedrock/Client` objects.  However, I really hope that we haven't been doing that -- it really makes more semantic sense to just create a single `Bedrock/Client` object and then reuse it.

cead22:
> Thoughts on this dramatically simpler solution?

Soudns good to me and it doesn't seem like we need to keep track of sockets per pids, host names or clusters

> However, I really hope that we haven't been doing that -- it really makes more semantic sense to just create a single Bedrock/Client object and then reuse it.

Looks like we are, for instance it seems like all workers are instantiated with an instance of a BR client objects [here](https://github.com/Expensify/Bedrock-PHP/blob/8689b0f/bin/BedrockWorkerManager.php#L237) but then inside the worker we instantiate a new instance instead of using the one passed during instantiation, for example in policy harvester [here](https://github.com/Expensify/Web-Expensify/blob/b250bb3e72/script/bwm/PolicyHarvester.php#L169). There are other workers that do this as well but I don't know the full extent of this.

quinthar:
> There are other workers that do this as well but I don't know the full extent of this.

Ah, well hopefully this isn't common.  I'd much rather do our state sharing explicitly -- by actually reusing the same object -- than implicitly (using a bunch of secret global variables).  It's unexpected/confusing to do `new Blah();` if all the state inside `Blah` is global -- in that case we should just have a global `$g_blah` variable or a more formal singleton pattern to make it more overt.  (And I recognize that the `Client` objects do share a blacklist between threads, but that seems a very different kind of data than live sockets.)

Anyway, so my vote would be to:

1. change [this line](https://github.com/Expensify/Bedrock-PHP/blob/master/src/Client.php#L331) to avoid re-connecting if we already have a socket (I think it was originally implemented this way actually)
2. Update [here](https://github.com/Expensify/Bedrock-PHP/blob/master/bin/BedrockWorkerManager.php#L216) to close the socket when forking
3. Unwind any extraneous `new Bedrock\Client()` calls and instead go toward explicit object reuse.

Thoughts?

cead22:
👍

iwiznia:
That proposal has several shortcomings: 
1. If we did that, then BWM won't be reusing sockets, which was what prompted this change in the first place. We would call getJobs on one socket, then close the socket when we fork and would have to reopen the socket again for our next call to getJobs, making the change moot.
2. The way we use the Client object (both for talking to webrock and auth) is not by instantiating the object once and then reusing it everywhere, we instantiate a new object when we need it, so in order to get benefits of reusing sockets everywhere, we'd need to rewrite all the usages of it, which is a much bigger and harder change.
3. Even if we moved to a singleton pattern or a global variable, we'd need to track the process that's using it there (in order for 2 threads to not interfere with each other), so we would just be moving this complexity to a different place (and probably duplicating it, because we would need it for auth and webrock and in both web and secure). This way, it keeps the complexity transparent from the consumers and just in one place.

iwiznia:
Regarding 1, I talked with David via slack and I misunderstood what he meant, we want to clear the variable once we fork, not close the socket. Still other 2 points apply.

quinthar:
> Even if we moved to a singleton pattern or a global variable, we'd need to track the process that's using it there (in order for 2 threads to not interfere with each other), so we would just be moving this complexity to a different place (and probably duplicating it, because we would need it for auth and webrock and in both web and secure). This way, it keeps the complexity transparent from the consumers and just in one place.

I think this is inaccurate.  When you fork a process, both processes have identical *copies* of all memory -- but they are *different* copies.  So after the fork, both BWM and the worker have different albeit identical copies of all the data.  As discussed, if the worker just clears (but does *not* close) its socket after being forked, then BWM is able to keep using its socket just fine -- and the worker will establish its own new socket when needed.  Given this, do you agree there is no multi-threaded concern so long as the worker clears its socket after forking?

> and probably duplicating it, because we would need it for auth and webrock and in both web and secure

I'm not sure what you mean.  These 2 tiny changes would only happen inside Bedrock-PHP, which is a shared resource.  So I don't think there is any duplication of code anywhere.  Do you agree?

iwiznia:
ok, I see where you are going to, I think I can simplify this a bit by not saving the sockets per pid. We still need the by cluster name grouping though.

iwiznia:
ok, updated, check it out.

iwiznia:
ok, check it out now. Attached is the web PR with all the replacements I needed to add to our current code. It wasn't an automatic find&replace, but wasn't that bad either.

src/Client.php
```diff
@@ -135,10 +140,31 @@ public function __construct(array $config = [])
         }
     }
 
-    public function __destruct()
+    /**
+     * Returns an instance of this class for the specified configuration. It will return the same instance if the same
+     * configuration was passed, unless clearInstancesAfterFork is called.
+     */
+    public static function get(array $config = []): Client
```
flodnv: As mentioned 1:1, I think a better name like `getInstance` would make more sense

src/Client.php
```diff
@@ -262,13 +288,19 @@ public function call($method, $headers = [], $body = '')
         while (!$response && count($hostConfigs)) {
             reset($hostConfigs);
             $numRetriesLeft = count($hostConfigs) - 1;
-            $hostName = key($hostConfigs);
-            $this->lastHost = $hostName;
+
+            // If we already have a socket for this pid and cluster, then we first try to reuse it
```
flodnv: `- for this pid`

src/Client.php
```diff
@@ -135,10 +140,31 @@ public function __construct(array $config = [])
...
+    {
+        $config = array_merge(self::$defaultConfig, $config);
+        ksort($config);
+        $hash = sha1(print_r($config, true));
+        if (isset(self::$instances[$hash])) {
+            return self::$instances[$hash];
+        }
+        $instance = new Client($config);
+        self::$instances[$hash] = $instance;
+
```
quinthar: Remove this blank line?

iwiznia: Ummm, the style checker should be checking this... @tgolen seems like your change is not working correctly?

iwiznia:
ok, off hold and ready to merge. Phan errors are a phan bug.

src/Client.php
flodnv: Indeed, fixing the style checker is out of scope of this PR.

PR #62, opened 2017-11-20
Change BWM queries to work more effectively

mattabullock:
Two things here:

1. Change it to use microtime instead of SQL DATETIME - datetime has terrible accuracy, only down to the second. Most of our jobs take less than a second, so that's not super helpful.
2. Calculate the average time outside of SQL - Turns out when you limit by target but do a SUM in sqlite, it still sums everything, disregarding the limit completely. 

# Tests
1. Queue tons of jobs
2. Check DB
3. Verify jobs are now shown as floats in the db
```
sqlite> SELECT started, ended FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT 10;
1511223595.47|1511223596.49
1511223595.29|1511223596.31
1511223595.44|1511223595.47
...
```
4. Verify jobs still queue correctly.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -68,7 +68,6 @@ if ($enableLoadHandler) {
     $localDB->open();
     $query = 'CREATE TABLE IF NOT EXISTS localJobs (
         localJobID integer PRIMARY KEY AUTOINCREMENT NOT NULL,
-        pid integer NOT NULL,
         jobID integer NOT NULL,
         jobName text NOT NULL,
         started text NOT NULL,
@@ -203,6 +202,13 @@ try {
                         'workerFileName' => $workerFilename,
                     ]);
 
+                    $localJobID = 0;
+
+                    // Open the DB connection after the fork in the child process.
+                    if ($enableLoadHandler) {
+                        $localDB->write("INSERT INTO localJobs (jobID, jobName, started) VALUES ({$job['jobID']}, '{$job['name']}', ".microtime(true).");");
+                        $localJobID = $localDB->getLastInsertedRowID();
+                    }
                     // Close DB connection before forking.
                     if ($enableLoadHandler) {
                         $localDB->close();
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -240,7 +240,7 @@
                             // Open the DB connection after the fork in the child process.
                             if ($enableLoadHandler) {
                                 $localDB->open();
-                                $localDB->write("INSERT INTO localJobs (pid, jobID, jobName, started) VALUES ($childPID, {$job['jobID']}, '{$job['name']}', DATETIME('now'));");
+                                $localDB->write("INSERT INTO localJobs (pid, jobID, jobName, started) VALUES ($childPID, {$job['jobID']}, '{$job['name']}', ".microtime(true).");");
```
iwiznia: Do we care that this column is defined as text in the schema?

bin/BedrockWorkerManager.php
```diff
@@ -273,7 +273,7 @@
                                 $jobs->failJob($job['jobID']);
                             } finally {
                                 if ($enableLoadHandler) {
-                                    $localDB->write("UPDATE localJobs SET ended=DATETIME('now') WHERE localJobID=$localJobID;");
+                                    $localDB->write("UPDATE localJobs SET ended=".microtime(true)." WHERE localJobID=$localJobID;");
```
iwiznia: Do we care that this column is defined as text in the schema?

bin/BedrockWorkerManager.php
mattabullock: I changed it to real for testing and it didn't change the results... not sure it matters.

iwiznia: ok

bin/BedrockWorkerManager.php
```diff
@@ -360,9 +360,10 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
     }
 
     // Calculate the speed of the last 2 batches to see if we're speeding up or slowing down
-    $oldBatchTimes = $localDB->read("SELECT CAST(STRFTIME('%s', ended) - STRFTIME('%s', started) AS int) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
+    $oldBatchTimes = $localDB->read("SELECT ended - started FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target OFFSET $target;");
     $oldBatchAverageTime = array_sum($oldBatchTimes) / count($oldBatchTimes);
-    $newBatchAverageTime = $localDB->read("SELECT SUM(CAST(STRFTIME('%s', ended) - strftime('%s', started) AS int)) / count(*) FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target;")[0];
+    $newBatchTimes = $localDB->read("SELECT ended - started FROM localJobs WHERE ended IS NOT NULL ORDER BY ended DESC LIMIT $target;");
+    $newBatchAverageTime = array_sum($newBatchTimes) / count($newBatchTimes);
```
quinthar: Or you could just do this as:
```
$oldBatchAverageTime = $localDB->read("
    SELECT SUM(time)/COUNT(*) FROM (
        SELECT ended-started 
        FROM localJobs 
        WHERE ended IS NOT NULL 
        ORDER BY ended 
        DESC LIMIT $target 
        OFFSET $target)
    )
");
```

That would keep the calculation fully inside SQL.

PR #64, opened 2017-11-28
Add support for 10x load testing

tylerkaraszewski:
@iwiznia 

Please review this change for 10x testing support. Please tell me if I need to change a version number somewhere.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -139,8 +139,8 @@ class Client implements LoggerAwareInterface
         $this->maxBlackListTimeout = $config['maxBlackListTimeout'];
 
         // If the caller explicitly set `mockRequests`, use that value.
-        if (isset($config['mockReqests'])) {
-            $this->mockRequests = $config['mockReqests'];
+        if (isset($config['mockRequests'])) {
+            $this->mockRequests = $config['mockRequests'];
         } else {
             // otherwise pull from the request headers.
             $requestHeaders = getallheaders();
@@ -199,7 +199,6 @@ class Client implements LoggerAwareInterface
             'stats' => new NullStats(),
             'writeConsistency' => 'ASYNC',
             'maxBlackListTimeout' => 1,
-            'mockReqests' => false,
         ], self::$defaultConfig, $config);
     }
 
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -103,17 +108,17 @@ class Client implements LoggerAwareInterface
...
-     * array|null           mainHostConfigs     List of hosts to attempt first
-     * array|null           failovers           List of hosts to use as failovers
-     * int|null             connectionTimeout   Timeout to use when connecting
-     * int|null             readTimeout         Timeout to use when reading
-     * LoggerInterface|null logger              Class to use for logging
-     * StatsInterface|null  stats               Class to use for statistics tracking
-     * string|null          writeConsistency    The bedrock write consistency we want to use
-     * int|null             maxBlackListTimeout When a host fails, it will blacklist it and not try to reuse it for up
-     *                                          to this amount of seconds.
+     *                      string               clusterName         Name of the bedrock cluster. This is used to separate requests made to
```
iwiznia: Weird spaces

src/Client.php
```diff
@@ -130,6 +135,17 @@ public function __construct(array $config = [])
         $this->writeConsistency = $config['writeConsistency'];
         $this->maxBlackListTimeout = $config['maxBlackListTimeout'];
 
+        // If the caller explicitly set `mockRequests`, use that value.
+        if (isset($config['mockReqests'])) {
```
iwiznia: Instead of checking if isset, add the `mockRequest` param with default `false` as the rest of params, it will also make `$this->mockRequests` always a boolean (now it's null if mockRequests isn't passed). Also, please add the config option to the phpdoc.

tylerkaraszewski:
@iwiznia - updated. Because I don't know what I'm doing, version goes from 1.1.1 to 1.1.5 to account for all my failed attempts.

flodnv:
Any tests? 😕

tylerkaraszewski:
Tests will be in the Web-Expensify change, which I understand isn't ideal.

flodnv:
Understood, but that doesn't prevent you from testing this..

On Wed, Nov 29, 2017 at 6:36 PM, Tyler Karaszewski <notifications@github.com

tylerkaraszewski:
Along with the auth/web changes that go alongside this, I've tested this in my VM. You need a working auth with mockRequest support and a working Web-Expensify with mockRequest support to easily test it, though.

If you check out all three branches, then you can enable mock requests via `www-dev-config.php` and see that this works.

PR #65, opened 2017-11-29
Fix retrying in failovers and BWM

iwiznia:
Fixes https://github.com/Expensify/Expensify/issues/66949


flodnv:
The 2 Phan errors are to be ignored as they're caused by a bug in Phan.

PR #71, opened 2017-12-05
Remove unused exceptions

mea36:
Fixes https://github.com/Expensify/Expensify/issues/65018 and https://github.com/Expensify/Expensify/issues/67763

**Tests**
**Timeout exception**
```
>>> try {$db->query('select 1 from table;');} catch (Exception $e) {}
>>> $e
=> Expensify\Bedrock\Exceptions\BedrockError {#403
     #message: "555 Timeout peeking command",
     #code: 555,
     #line: 71,
   }
```
**Query failed exception**
```
>>> try {$db->query('select;');} catch (Exception $e) {}
>>> $e
=> Expensify\Bedrock\Exceptions\BedrockError {#404
     #message: "502 Query failed - near ";": syntax error",
     #code: 502,
     #line: 67,
   }
```


src/DB.php
```diff
@@ -66,11 +64,11 @@ public function run(string $sql, bool $idempotent, int $timeout = 60000000): Res
         ));
 
         if ($response->getCode() === self::CODE_QUERY_FAILED) {
```
mea36: I don't know if we want to keep this here anyway.  I'm assume we want to remove it but asking nonetheless

iwiznia: I think we can remove it, the error would say query failed anyway, right?

mea36:
@iwiznia updated
I removed `CODE_QUERY_FAILED` and updated the exception to have a better message and include the response code

I'll update the version after you approve this

mea36:
Also, I found a bug, check it out https://github.com/Expensify/Expensify/issues/67763

iwiznia:
As a general rule, please add hold or wip to the name of PRs in this repo, since we need to create a new version and we don't want to merge it by mistake before we create the version.

src/DB.php
mea36: I added this back so we can have an exception message that includes the error

src/Client.php
```diff
@@ -601,7 +601,7 @@ private function extractResponseHeaders(array $responseHeaderLines)
         $responseHeaders = [];
         foreach ($responseHeaderLines as $responseHeaderLine) {
             // Try to split this line
-            $nameValue = explode(":", $responseHeaderLine);
+            $nameValue = explode(":", $responseHeaderLine, 2);
```
quinthar: Ooh, nice catch!

src/DB.php
quinthar: I don't follow this -- why don't we call `$response->getError()` on all non-200 responses?  Why is 502 special?

mea36:
Updated and added tests.  I'll add the version when this is approved

src/DB.php
mea36: `error` is only added when we throw a 502

https://github.com/Expensify/Bedrock/blob/ce238f0/plugins/DB.cpp#L78-L83

quinthar: Ah, makes sense.  thanks!

src/Client.php
```diff
@@ -10,6 +10,8 @@
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 
+define('TRAVIS_RUNNING', isset($_SERVER['TRAVIS_COMMIT']));
```
iwiznia: Why did you add this? I don't think we can do this, at least you need to check that the constant is not defined before.

mea36: Sorry for the delay, this fell off of my radar.
I did this because phan was yelling at me https://travis-ci.org/Expensify/Bedrock-PHP/builds/313619228

But you're right, I should check if this is defined first

mea36: Actually, after taking a 2nd look at the code, I think I should add a comment for phan to ignore the error
```
FAIL! src/Client.php: src/Client.php:436 PhanUndeclaredConstant Reference to undeclared constant \TRAVIS_RUNNING
FAIL! src/Client.php: src/Client.php:640 PhanUndeclaredConstant Reference to undeclared constant \TRAVIS_RUNNING
```

mea36:
@iwiznia updated.  I'll add a version when you approve these changes

mea36:
@iwiznia updated

PR #72, opened 2017-12-07
[hold]Clear stuck jobs on BWM restart

mattabullock:
We're continuously getting jobs that get stuck running and never update. The best way to deal with that from the standpoint of Bedrock-PHP is to clear out stuck jobs whenever it is restarted, since this happens frequently enough that we shouldn't ever gather too many stuck jobs.

Hold for packaging.

# Issue
https://github.com/Expensify/Expensify/issues/61961

# Tests
1. Insert jobs into DB with `ended=null`
2. Start bwm with `--enableLoadHandler`
3. Check that the initial jobs were deleted.


JLiebsExpensify:
Rather than counting on when it's restarted, perhaps we just automatically
expire any job after 1 hour? It feels like the restart frequency of bwm is
an important factor, so why not program that in explicitly rather than
living it up chance.

quinthar:
Wow I have no idea why it claims that Josh was providing that comment. However the comment still stands, it feels strange to rely upon the arbitrary frequency of restarts to do something that we have concluded is actually quite important. If it's important, let's just choose the right frequency to do it and then program in that frequency.

mattabullock:
Updated.

mattabullock:
Wow, thanks for catching that, bad testing on my part.

mattabullock:
How long do we think is safe to delete after? an hour?

JLiebsExpensify:
Well what is the retryAfter set to?  We should just match that

JLiebsExpensify:
And we might also want to set the max execution time of a worker to that
just to be safe

mattabullock:
Do you mean to set the purge time on a job by job basis?

I asked and retryAfter isn't set for most jobs, so is that a good option?

quinthar:
Actually thinking more on this, I'm not sure why this is a problem, because stuck jobs "cost" nothing, and thus AIMD should correctly recognize that we can safely add more jobs.  To be clear, stuck jobs are a problem (and I'm curious why we are having so many of them), but I'm questioning whether they actually matter to the problem at hand.

cead22:
What I theorized is that because these jobs have no `ended` value then [this query](https://github.com/Expensify/Bedrock-PHP/blob/e4b5d20/bin/BedrockWorkerManager.php#L368) returns it and thus they always "count" as a running job so if we have 1 of these and our target is 10 we'll always create at most 9 more jobs because the stuck one will be assumed to be running/active.

Yesterday ran that same query with an extra condition `AND created > december_5` and got results back (note that we tested `december_5` in the date format and epoch format and got results on both queries)

quinthar:
Yes, I understand that. But the design of the algorithm is that we keep
adding jobs so long as things don't slow down. A job that is stuck is not
consuming any cost, and thus can't slow anything down. So whether we have
one stuck job or a million stuck jobs, the algorithm should still increase
the Target because none of them cost anything.

cead22:
Ah yeah good point. I just re-read the code and checked the logs and the target is going up so even though we're not hitting the actual target (when we have stuck jobs) we are increasing the number of jobs being run at a time.

mattabullock:
I think that's fine in theory, but in practice it's not necessarily true. If you could prove that the target is always 2 (or 5 or even 9) higher, then I'd agree with you.

It also has the potential to completely shut down BWM altogether if the number of stuck jobs reaches a point that it's higher than the target.

mattabullock:
I'm going to attempt to push this one through again. We've had multiple times where the server died and when it started up again it was unable to run any jobs for 10+ hours (because I was offline and didn't look at the graphs). I think for this specific case it makes sense to just clear open jobs on startup since that seems to be the most common case for where we're seeing failures.

quinthar:
Hm, I'm still not sure why we don't just expire them when their retryAfter expires?  Basically, every once in a while delete all jobs that we know Bedrock has given up on because they have been running longer than their expiration.

That said, the scenario where we have more stuck jobs than the target -- none of which have any retryAfter -- is pretty extreme, and should also be solved.  But solving it on restart feels like a super hacky way to do it that will just kick the underlying problems down the road.  In practice, even if a job lacks retryAfter, we're comfortable blowing it away after an hour or so.

So what about just deleting stuck jobs after their max(1 hour, retryAfter) expires?

iwiznia:
> even if a job lacks retryAfter

Most jobs don't have retryAfter, the only ones that are in use that have it are the concierge related ones.

> So what about just deleting stuck jobs after their max(1 hour, retryAfter) expires?

Clearing it from the real DB or from the local DB used by AIMD? If it's the former, then I'm against it, because we need the job there to investigate why it got stuck, if it's the latter I'm fine with it.

quinthar:
Correct, I mean the latter -- just clear out of the AIMD database after an hour (or the retryAfter, if set).  And again, I think that's solving for a pretty extreme edge case, so I am fine enduring up to an hour in that case personally.  Really, we shouldn't have that many stuck jobs, and if we do, *that* is the problem we should be focused on fixing.

iwiznia:
I'm fine with that too.
To clarify, this isn't about stuck jobs really (at least not what prompted us to revive this thread). The problem was that the web node died due to OOM, killing anything that was running in it. So the stuck jobs were the consequence, not the cause.

mattabullock:
@iwiznia I'm reopening this because I was asked to a few times

PR #75, opened 2017-12-13
Add timers for different parts of BWM process

mattabullock:
Hold for packaging.

I added timers that get sent to StatsD to help us evaluate how well BWM is performing.

Related: https://github.com/Expensify/Expensify/issues/61961

#### Timers added
* Full loop (excluding GetJobs)
* CheckVersionFile
* GetNumberOfJobsToQueue
* Insert into localJobsDB
* Read from localJobsDB
* Delete from localJobsDB
* Update localJobsDB
* Fork
* Opening the DB
* Closing the DB

#### Counters added
* Target

### Tests
1. Run BWM with --enableLoadHandler and make sure it runs with no warns or errors.

No QA

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -107,8 +107,8 @@ try {
         $iteration++;
         $logger->info("Loop iteration", ['iteration' => $iteration]);
 
-        $isFirstTry = true;
         // Step One wait for resources to free up
+        $isFirstTry = true;
         while (true) {
             $childProcesses = [];
             // Get the latest load
@@ -138,6 +138,7 @@ try {
             }
         }
 
+        // Check to see if BWM was able to get jobs on the first attempt. If not, it would add a full second each time it failed, skewing the timer numbers.
         if ($isFirstTry) {
             $stats->timer("bedrockWorkerManager.fullLoop", microtime(true) - $loopStartTime);
         }
@@ -164,9 +165,8 @@ try {
             }
         }
 
-        $loopStartTime = microtime(true);
-
         // Found a job
+        $loopStartTime = microtime(true);
         if ($response['code'] == 200) {
             // BWM jobs are '/' separated names, the last component of which
             // indicates the name of the worker to instantiate to execute this
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -58,12 +58,13 @@ $target = $minSafeJobs;
 // Configure the Bedrock client with these command-line options
 $bedrock = Client::getInstance($options);
 
-// Prepare to use the host logger, if configured
+// Prepare to use the host logger and stats client, if configured
+$stats = $bedrock->getStats();
 $logger = $bedrock->getLogger();
 $logger->info('Starting BedrockWorkerManager', ['maxIterations' => $maxIterations]);
 
 // Set up the database for the AIMD load handler.
-$localDB = new LocalDB($pathToDB, $logger);
+$localDB = new LocalDB($pathToDB, $logger, $stats);
 if ($enableLoadHandler) {
     $localDB->open();
     $query = 'CREATE TABLE IF NOT EXISTS localJobs (
@@ -84,7 +85,6 @@ $versionWatchFileTimestamp = $versionWatchFile && file_exists($versionWatchFile)
 
 // Wrap everything in a general exception handler so we can handle error
 // conditions as gracefully as possible.
-$stats = $bedrock->getStats();
 try {
     // Validate details now that we have exception handling
     if (!is_dir($workerPath)) {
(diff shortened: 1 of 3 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -124,20 +126,26 @@
...
+                $stats->counter('bedrockWorkerManager.targetJobs', $target);
                 break;
             } else {
                 $logger->info('Not safe to start a new job, waiting 1s and trying again.', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
+                $isFirstTry = false;
                 sleep(1);
             }
         }
 
+        if ($isFirstTry) {
```
iwiznia: Not sure I understand this... why do we do this? Also, isn't it tracking basically the same as `getNumberOfJobsToQueue`?

bin/BedrockWorkerManager.php
```diff
@@ -210,7 +220,7 @@
                         $localDB->close();
                     }
                     pcntl_signal(SIGCHLD, SIG_IGN);
-                    $pid = pcntl_fork();
+                    $pid = $stats->benchmark('bedrockWorkerManager.fork', function () { return pcntl_fork(); });
```
iwiznia: Huh?? Is this useful? I thought forking was immediate...

src/LocalDB.php
```diff
@@ -22,23 +22,32 @@ class LocalDB
...
     }
 
     /**
      * Opens a DB connection.
      */
-    public function open() {
+    public function open()
+    {
         if (!isset($this->handle)) {
+            $startTime = microtime(true);
```
iwiznia: Why not use benchmark?

src/LocalDB.php
```diff
@@ -48,8 +57,10 @@ public function open() {
     public function close()
     {
         if (isset($this->handle)) {
+            $startTime = microtime(true);
```
iwiznia: Same, why not benchmark?

src/LocalDB.php
```diff
@@ -61,7 +72,8 @@ public function close()
     public function read(string $query)
     {
         $result = null;
-        while(true) {
+        $returnValue = null;
```
iwiznia: Given that you added this, I think you can remove the `??` in  line 94, right? Since the variable will always be defined.

bin/BedrockWorkerManager.php
```diff
@@ -106,6 +107,7 @@
         $iteration++;
         $logger->info("Loop iteration", ['iteration' => $iteration]);
 
+        $isFirstTry = true;
```
quinthar: Can you move this underneath the comment please?

bin/BedrockWorkerManager.php
quinthar: At the very least can you provide a comment above this line that explains what is going on here?

bin/BedrockWorkerManager.php
```diff
@@ -156,6 +164,8 @@
             }
         }
 
+        $loopStartTime = microtime(true);
```
quinthar: Can you move this beneath the comment below?

quinthar:
Once we're done with this timing, can we please please please rip all this code back out?  This timing junk has polluted our Auth PHP code for years for virtually no value -- once we have the answers we need, let's clean this back out and keep the codebase tidy.

bin/BedrockWorkerManager.php
mattabullock: @flodnv sent me this, so I figured it would be worth checking.

https://blog.famzah.net/2009/11/20/fork-gets-slower-as-parent-process-use-more-memory/

src/LocalDB.php
mattabullock: Woah weird, I added this line because PHPStorm complained but now it's not complaining. I'm just gonna remove it.

mattabullock:
>Once we're done with this timing, can we please please please rip all this code back out? This timing junk has polluted our Auth PHP code for years for virtually no value -- once we have the answers we need, let's clean this back out and keep the codebase tidy.

That's fine with me.

mattabullock:
updated

src/LocalDB.php
```diff
@@ -79,15 +92,15 @@ public function read(string $query)
             $returnValue = $result->fetchArray(SQLITE3_NUM);
         }
 
-        return $returnValue ?? null;
+        return $returnValue ?? [];
```
iwiznia: You sure about this change? It's a different behavior than before and I'm not sure why it needed changing.

mattabullock: I realized the actual reason I changed it was because phan was failing.

iwiznia: Well, can you revert it back? This changes the behavior of this method, you can make phan pass without changing the behavior.

mattabullock: Yeah... I just hate that the value can be an array, false, or null.

iwiznia: I then realized this is `LocalDB` and not `DB` and we are not really using the `null` values anywhere 🤦‍♂️

PR #77, opened 2018-01-17
Reorganize CI-related repository files

fnwbr:
This is in order to resolve https://github.com/Expensify/Web-Expensify/pull/19896 -
Changes do not affect any functionality or features of the actual Bedrock-PHP code


composer.json
```diff
@@ -26,7 +26,7 @@
     "autoload": {
         "psr-4": {
             "Expensify\\Bedrock\\": "src",
-            "Expensify\\Libs\\": "lib/src"
+            "Expensify\\BedrockLibs\\": "lib/src"
```
iwiznia: Wait, really?? You can do this in code and it works??? You can do `use Expensify\Libs\ArrayUtils` in web-e?

iwiznia: Wait I just realized you changed _some_ of them, not all... so

lib/src/CommandLine.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Expensify\Libs;
+namespace Expensify\BedrockLibs;
```
iwiznia: ok, I see what happened, we copied this files, but we didn't change the namespace... we should've changed the namespace to be `Expensify\Bedrock` like all the rest of files in this repo. I don't like us having 2 namespaces in this same repo.

fnwbr: Those files in this namespace don't really belong to the project code and are merely helpers... are you sure you want them to be moved to `Expensify\Bedrock` and therefore into `src/`?
If so, what do you want them to be namespaced under?

iwiznia: Talked offline. We'll keep the namespace but put it under `Expensify\Bedrock`.

fnwbr:
Updated

PR #78, opened 2018-01-22
[HOLD discussion] Replace bad ascii with characters rather than deleting them.

johnmlee101:
cc: @coleaeason 

Before we would purge bad ascii when receiving requests. I tested this with our `#001private_amountOwed` NVP stored in the DB. We shouldn't purge bad ascii or else we change it to `private_amountOwed` which then is valid... which is bad. 

Right now I set it to replace the bad ascii to the ascii code itself, prepended with a `#`. So `#001private_amountOwed` => `#1private_amountOwed`


johnmlee101:
Closing this out, as we want a better upstream solution. Deleting the tag as well.

PR #83, opened 2018-03-07
Protect against sql injecting ourselves

mattabullock:
Emails have special characters in them, and it was allowing us to sql inject ourselves in the BWM db.

Related to: https://github.com/Expensify/Expensify/issues/66022

HOLD for updating version

# Tests
1. Queue job with special characters in the params (i.e. `www-prod/TestJob?asdf=''''`)
2. Run BWM
3. Confirm that it doesn't crash and is inserted into the db effectively.
```
3|66|www-prod/TestJob?asdf=''''f'|1520450955.5|1520450955.62
```

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -194,11 +194,10 @@ try {
             $jobsToRun = $response['body']['jobs'];
             foreach ($jobsToRun as $job) {
                 $localJobID = 0;
-                $safeJobID = SQLite3::escapeString($job['jobID']);
                 $safeJobName = SQLite3::escapeString($job['name']);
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -194,8 +194,11 @@
             $jobsToRun = $response['body']['jobs'];
             foreach ($jobsToRun as $job) {
                 $localJobID = 0;
+                $safeJobName = SQLite3::escapeString($job['name']);
```
iwiznia: NAB: this could be inside the if.

mattabullock:
ready to merge

PR #84, opened 2018-03-08
Echo info about the state of jobs while in debug mode

mea36:
I find it ridiculously helpful to see the status of the jobs that the bwm is picking up when working in my dev env.

This adds `--debugLogs` as a command line argument when running the script.  If this is set, it echos the timestamp, action, job name, and requestID

I'm not sure how everyone feels about, I'll update the version number if everyone is 👍


iwiznia:
I don't like this, mainly because we already print that information in the logs, and if any piece of information is not there, then you can easily add it and if it's too noisy you can add a `debug` log line so it only shows on dev. Why can't you use the logs to see this information?

mea36:
The logs are way too noisy.  I didn't think of adding a `debug` line.  But even if I did that I'd have to create a script to cut out parts of the log that I don't need (like the filename).

So I'll just write a script for this

PR #85, opened 2018-03-22
Instrument cache

iwiznia:
@fnwbr please review.

Tests:
- We have some automated tests in the web repo.
- Tested:
```
➜  Web-Expensify git:(staging) ✗ ../script/psysh.sh
Psy Shell v0.8.1 (PHP 7.0.28-1+ubuntu14.04.1+deb.sury.org+1 — cli) by Justin Hileman
New version is available (current: v0.8.1, latest: v0.8.17)
>>> $c = new Expensify\Bedrock\Cache(Expensify\Bedrock\Client::getInstance())
=> Expensify\Bedrock\Cache {#388}
>>> $c->write('something', '["uno"]')
=> [
     "headers" => [
       "Content-Length" => "0",
     ],
     "body" => [],
     "size" => 46,
     "codeLine" => "202 Successfully queued",
     "code" => 202,
   ]
>>> $c->read('something')
=> [
     "headers" => [
       "commitCount" => "19",
       "name" => "something/",
       "nodeName" => "bedrock",
       "peekTime" => "3764",
       "totalTime" => "3874",
       "unaccountedTime" => "80",
       "Content-Length" => "7",
     ],
     "body" => [
       "uno",
     ],
     "size" => 139,
     "codeLine" => "200 OK",
     "code" => 200,
   ]
```

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -34,13 +34,7 @@ class Cache extends Plugin
             'key' => $name,
             'version' => $version,
         ]);
-
-        return $this->call([
-            "ReadCache",
-            [
-                "name" => $fullName,
-            ],
-        ]);
+        return $this->call("ReadCache", ["name" => $fullName]);
     }
 
     /**
@@ -71,21 +65,16 @@ class Cache extends Plugin
             $headers["name"] = "$name/";
         }
 
-        return $this->call([
-                "WriteCache",
-                $headers,
-                $value,
-        ]);
+        return $this->call("WriteCache", $headers, $value);
     }
 
     /**
      * Call the bedrock cache methods, and handle connection error.
      *
-     * @param array $parameters
-     *
+     * @param string $body
      * @return mixed|null
      */
-    private function call(array $parameters)
+    private function call(string $method, array $headers, $body = '')
     {
         if (self::$hasFailed) {
             $this->client->getLogger()->info('Skip Bedrock Cache call because we have failed before');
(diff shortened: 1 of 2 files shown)


src/Cache.php
```diff
@@ -71,21 +65,17 @@ public function write($name, $value, $version = null)
...
     /**
      * Call the bedrock cache methods, and handle connection error.
      *
-     * @param array $parameters
+     * @param string $body
      *
      * @return mixed|null
      */
-    private function call(array $parameters)
+    private function call(string $method, array $headers, $body = '')
```
fnwbr: `string $body = ''` into the function head, and just get rid of the `@param` PHPdoc

src/Cache.php
```diff
@@ -34,13 +34,7 @@ public function read($name, $version = null)
...
             'version' => $version,
         ]);
-
-        return $this->call([
-            "ReadCache",
-            [
-                "name" => $fullName,
-            ],
-        ]);
+        return $this->call("ReadCache", ["name" => $fullName]);
```
fnwbr: style: newline before return (I think 😄)

src/Cache.php
iwiznia: I thought about it but decided not to as I would need to ensure all callers pass this as string.

src/Cache.php
iwiznia: We removed that not long ago 😄

src/Cache.php
fnwbr: When do we want to do it if not now?
Given that we only have a handful of usages really...

src/Cache.php
fnwbr: Damn... it's definitely like this for JS... ~love~ hate the consistency... anyways; NAB

PR #86, opened 2018-03-26
[wip]Check for mocked jobs and mark them as finished

mattabullock:


src/Jobs.php
```diff
@@ -217,6 +217,7 @@ public function getJobs(string $name, int $numResults): array
         $headers = [
             "name" => $name,
             "numResults" => $numResults,
+            "mockRequest" => true,
```
tylerkaraszewski: What does this do? are we sending it on every single request?

mattabullock:
Off hold

tylerkaraszewski:
Title still says hold.

mattabullock:
now really off hold

PR #90, opened 2018-04-20
Sleep after starting some jobs

iwiznia:
Fixes https://github.com/Expensify/Expensify/issues/77430

We use GetJobs to try to dequeue several jobs in the same call, but since we are calling it immediately after starting all jobs, we usually end up retrieving very few jobs per call because the jobs did not have time to finish yet. By adding a sleep we hope to make the GetJobs calls to retrieve more jobs per run, which in turn reduces the possibilities of the command conflicting.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -355,7 +355,7 @@ try {
                     $jobs->failJob($job['jobID']);
                 }
             }
-            // After starting a few jobs, wait for a couple of seconds to give time to some jobs to finish so its more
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -355,6 +355,9 @@
                     $jobs->failJob($job['jobID']);
                 }
             }
+            // After starting a few jobs, wait for a couple of seconds to give time to some jobs to finish so its more
```
cead22: Typo `it's`

mattabullock:
@cead22 has to approve before it can be merged.

PR #95, opened 2018-05-01
[HOLD]Only send mockRequest when it's specified

mattabullock:
Changing how we request mocked jobs.

Holding on https://github.com/Expensify/Bedrock/pull/449

## Tests
Integration tests on web work
Run BWM and it runs/closes both mocked jobs and normal jobs.


mattabullock:
Off hold

PR #97, opened 2018-05-03
Fix preferred host when host doesn't exist in the configs

coleaeason:
@mattabullock please review. This broke pinning hosts, aka setting the main/failover config to the same arrays, then having a preferred host come back that is not in those configs. We would then return a host with no port, which resulted in an error when the client tried to reconnect to the preferred host.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -521,7 +521,7 @@ class Client implements LoggerAwareInterface
         shuffle($failoverHostNames);
         $mainHostName = array_rand($this->mainHostConfigs);
         $preferredHost = in_array($preferredHost, $cachedHostConfigs) ? [$preferredHost] : [];
-        $hostNames = array_unique(array_merge([$preferredHost], [$mainHostName], $failoverHostNames));
+        $hostNames = array_unique(array_merge($preferredHost, [$mainHostName], $failoverHostNames));
 
         $nonBlackListedHosts = [];
         foreach ($hostNames as $hostName) {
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -520,7 +520,8 @@ private function getPossibleHosts(string $preferredHost = null)
         $failoverHostNames = array_keys($this->failoverHostConfigs);
         shuffle($failoverHostNames);
         $mainHostName = array_rand($this->mainHostConfigs);
-        $hostNames = array_unique(array_merge($preferredHost ? [$preferredHost] : [], [$mainHostName], $failoverHostNames));
+        $preferredHost = in_array($preferredHost, $cachedHostConfigs) ? [$preferredHost] : [];
```
iwiznia: This change is wrong, this is not doing what it was intended.... why are you changing this?? This has nothing to do with pinning hosts and it's a new constant added recently.

iwiznia: The correct fix for this is to ignore the preferred host if it's not in the configs... can you please update it?

iwiznia: Talked via slack, leaving it here for context: Sorry, my comment is crap... the `$cachedHostConfigs` is an array like: `['host' => [..], 'host2' => [...]` and `in_array` does not check if the key exists, checks if the value exists, so `$preferredHost` will always be set to `[]` (edited)
And WRT `this has nothing to do with pinning hosts` , I was wrong, I saw the problem in an issue later.
In hindsight, `$preferredHost` should've been named `$preferredHostName`.

PR #98, opened 2018-05-07
Check if the preferred host exists in keys not in array

coleaeason:
@iwiznia please review. Resolves https://github.com/Expensify/Bedrock-PHP/pull/97#discussion_r186145280

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -520,7 +520,7 @@ class Client implements LoggerAwareInterface
         $failoverHostNames = array_keys($this->failoverHostConfigs);
         shuffle($failoverHostNames);
         $mainHostName = array_rand($this->mainHostConfigs);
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -520,7 +520,7 @@ private function getPossibleHosts(string $preferredHost = null)
         $failoverHostNames = array_keys($this->failoverHostConfigs);
         shuffle($failoverHostNames);
         $mainHostName = array_rand($this->mainHostConfigs);
-        $preferredHost = in_array($preferredHost, $cachedHostConfigs) ? [$preferredHost] : [];
+        $preferredHost = array_key_exists($preferredHost, $cachedHostConfigs) ? [$preferredHost] : [];
```
iwiznia: Phan is complaining about this because you can't pass `null` to `array_key_exists`

iwiznia:
Bump! I want this to be working to get data to check the performance implication of the preferredHost.

coleaeason:
This is ready

PR #99, opened 2018-06-30
Add RUNQUEUED constant

pecanoro:
I am using the constant here https://github.com/Expensify/Web-Expensify/pull/21535/files#diff-1b31bee8fdcfb3a3a71485dcfb65511eR60

### DIFF
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -31,6 +31,13 @@ class Jobs extends Plugin
     const STATE_RUNNING = "RUNNING";
 
     /**
+     * State of a job that is currently in the runqueued state.
+     *
+     * @var string
+     */
+    const STATE_RUNQUEUED = "RUNQUEUED";
+
+    /**
      * State of a job that is currently queued.
      *
      * @var string
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -32,10 +32,8 @@ class Jobs extends Plugin
 
     /**
      * State of a job that is currently in the runqueued state.
-     *
-     * @var string
      */
-    const STATE_RUNQUEUED = "RUNQUEUED";
+    const STATE_RUNQUEUED = 'RUNQUEUED';
 
     /**
      * State of a job that is currently queued.
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "1.4.3",
+    "version": "1.4.4",
     "authors": [
         {
             "name": "Expensify",


src/Jobs.php
```diff
@@ -30,6 +30,13 @@ class Jobs extends Plugin
      */
     const STATE_RUNNING = "RUNNING";
 
+    /**
+     * State of a job that is currently in the runqueued state.
+     *
+     * @var string
```
quinthar: Out of curiosity, why the `@var string`?   Is this telling some system somewhere that `STATE_RUNQUEUED` is a string, and if so, which system?  I ask because in the style guide it seems to suggest that this kind of comment is extraneous and should be skipped, but it doesn't explicitly seem to cover the case of an untyped constant.  (Can we just specify the type of the constant in the code?)  I realize this conversation has probably happened a hundred times before, if so, sorry for missing it!

pecanoro: Yes, that was exactly my reasoning, but also true that a constant is not going to change its type and the type is pretty straightforward after seeing the assigned valued. Not sure what we want to do in this case. Any suggestions? @cead22

pecanoro:
I am going to reassign to @fnwbr because I think Ioni is on vacation.

src/Jobs.php
cead22: This tells our php code analyzers the type which I'm pretty sure it needs to make checks in the rest of code.

There's no way to specify this in the code itself unfortunately, so I suggest leaving the type annotation in.

quinthar: So if we remove that, we should be able to prove very immediately which code formatter needs it, because it'll puke, right?

pecanoro: I don't think it will puke, it simply won't run those extra checkings that our code analyzers run to make sure there are no code inconsistencies.

fnwbr: Following our own style guide I also think this shouldn't be here: https://github.com/Expensify/Style-Guide/blob/master/php.md#constants

At the same time that means the rest of the code isn't following it either. So it's really a chicken-and-egg type of problem here.

The only thing speaking for leaving this here is consistency with the rest of the code, I guess. Since it's a constant (and not a class property) all IDEs and static code analyzers will know that this is a string anyways.

src/Jobs.php
```diff
@@ -30,6 +30,13 @@ class Jobs extends Plugin
...
+     */
+    const STATE_RUNQUEUED = "RUNQUEUED";
```
fnwbr: Single quotes: https://github.com/Expensify/Style-Guide/blob/master/php.md#string

src/Jobs.php
cead22: If the static analyzers know the const is a string (which makes sense, I just didn't think about it) then I'm happy removing this.

pecanoro:
Updated!

src/Jobs.php
quinthar: Ya, let's remove it.  Let's not train everybody to follow rules that nobody knows the origin of, and nobody knows the value of -- especially when in this case, we've identified there is literally no value.

PR #100, opened 2018-09-26
AIMD2 2.0 Job Scheduling

tylerkaraszewski:
Fixes: https://github.com/Expensify/Expensify/issues/86281

There are a few major differences between AIMD 1.0 and AIMD 2.0. I'll summarize:

1. AIMD 2.0 uses a sliding time window (default 10 seconds) to determine the average length of time it takes for jobs to run. This gives it a lot more data to work with than just the last N jobs (where N was usually 10) to determine if jobs are getting faster or slower, but still should react to changes within a few seconds.
2. All of the constants (like the sliding window length of 10 seconds, the amount to backoff, the rate to increase by, etc) are command-line configurable so we can tweak this in production.
3. AIMD 2.0 adds a "blackout window" to prevent to successive loops from both backing off the target rate, allowing things to stabilize for a bit before dropping back a second time.
4. AIMD 2.0 has a djsuted defaults for some of the values that determine how fast we ramp up and down.

Additionally, there are two things that I'm looking at at bug fixes that may drastically improve effectiveness here.
1. AIMD 1.0 would *only* increase the target number of jobs if we already were running the exact number of jobs matching the current target. This meant that if our target was 10, and we were running 9 jobs, we would not increase the target. Because jobs can start and stop at any time, and we are never supposed to exceed the target number of jobs, this meant that most of the time, we wouldn't be running enough jobs to increase the target. AIMD 2.0 allows the target to increase well beyond the current number of running jobs, at it's specified rate. It will not increase beyond 2x the current number of running jobs (it also will not drop back to 2x the current number of running jobs just because jobs finish).
2. I've left a comment on the relevant line, but a query to clean up old jobs from over a minute ago seemed to clean up any jobs from earlier than a minute *in the future*, which may have given us quite bad data to average from.

Here's a screenshot of AIMD2.0 scheduling mock jobs in testing (the pink line is a moving average of currently running jobs):
screen shot 2018-09-26 at 3 33 47 pm

You can see it ramps up from 10 to ~70 simultaneous jobs and then stays more or less at that level, which is the intention.


bin/BedrockWorkerManager.php
```diff
@@ -128,15 +159,32 @@
...
+                // before retrying, which means all the current jobs can finish and leave nothing happening. Disabling
+                // it leads to dramatically higher job scheduling, at the cost of a potentially higher load on the DB,
+                // but that will be rectified in an upcoming change as well.
                 $logger->info('Safe to start a new job, checking for more work', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
                 $stats->timer('bedrockWorkerManager.numberOfJobsToQueue', $target);
                 $stats->timer('bedrockWorkerManager.targetJobs', $target);
                 break;
             } else {
-                $logger->info('Not safe to start a new job, waiting 1s and trying again.', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
-                $localDB->write('DELETE FROM localJobs WHERE started<'.(microtime(true) + 60 * 60).' AND ended IS NULL;');
```
tylerkaraszewski: The `+` instead of a `-` here seems like it would delete most of the data we used to average job times. @mattabullock said there was a problem where we saw "not enough data" loglines a lot more often than he expected, I assume this is what caused that.

tylerkaraszewski:
I suppose I have to change some composer files or something with this change, right?

agzakula:
Exciting!!

bin/BedrockWorkerManager.php
mattabullock: Wow, seems like a stupid mistake on my part... Wonder what it would look like with only this change...

mattabullock:
Yeah you'll have to make a web PR with the updated version after this is merged.

Also I thought we fixed the reply from email thing... @coleaeason ?

bin/BedrockWorkerManager.php
```diff
@@ -128,15 +159,32 @@
...
+                $jobsToQueue = getNumberOfJobsToQueue2();
+            } else {
+                $logger->warning('Invalid AIMD Version', ['aimdVersion' => $aimdVersion]);
+                exit(1);
+            }
+
+            if ($load > $maxLoad) {
+                $logger->info('Not safe to start a new job, load is too high, waiting 1s and trying again.', ['load' => $load, 'MAX_LOAD' => $maxLoad]);
+                sleep(1);
+            } elseif ($jobsToQueue > 0/*$minSafeJobs / 2*/) {
```
mattabullock: The reason we implemented this in the first place is because of a change where we wanted to begin using GetJobs instead of GetJob (multiple vs. singular). We made that change so we would only have to do one round trip on the network for every 5+ jobs instead of for every job. The issue with not putting a minimum to get was things finished quickly enough that we were routinely getting a single job with GetJobs, which isn't any more efficient than just running GetJob.

So I'd say (because of your comment) maybe we shouldn't wait a full second after getting less than the minSafeJobs and instead just immediately loop again?

bin/BedrockWorkerManager.php
```diff
@@ -458,6 +506,108 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
...
 }
 
+/**
+ * Determines whether or not we call GetJob and try to start a new job
+ *
+ * @return int How many jobs it is safe to queue.
+ */
+function getNumberOfJobsToQueue2(): int
+{
+    global $backoffThreshold,
```
mattabullock: Not sure how everyone feels about using globals here instead of passing them in as variables, I think that goes against our style (since if you made changes here they would perpetuate beyond this function call). So I'd say you should pass them through as params instead. Only reason we were doing that stupid redundant anonymous function was because of the benchmark I believe, so you could probably just pass them in normally.

bin/BedrockWorkerManager.php
```diff
@@ -458,6 +506,108 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
...
+    $lastIntervalAverage = floatval($temp0[1]);
+
+    $q1 = 'SELECT COUNT(*), AVG(ended - started) FROM localJobs WHERE ended IS NOT NULL AND ended > '.$twoIntervalsAgo.' AND ended < '.$oneIntervalAgo.';';
+    $temp1 = $localDB->read($q1);
+    $previousIntervalCount = $temp1[0];
+    $previousIntervalAverage = floatval($temp1[1]);
+
+    // Following block is only for testing.
+    if ($secondElapsed) {
+        echo "$secondElapsed, $numActive, $target, $lastIntervalAverage\n";
```
mattabullock: Probably want to use the built in logger for this because I don't think we don't do anything with the stdout of this program.

bin/BedrockWorkerManager.php
```diff
@@ -458,6 +506,108 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
...
+    $timeSinceLastRun = $now - $lastRun;
+    $lastRun = $now;
+    $oneIntervalAgo = $now - $intervalDurationSeconds;
+    $twoIntervalsAgo = $oneIntervalAgo - $intervalDurationSeconds;
+
+    // Look up how many jobs are currently in progress.
+    $numActive = intval($localDB->read('SELECT COUNT(*) FROM localJobs WHERE ended IS NULL;')[0]);
+
+    // Look up how many jobs we've finished recently.
+    $q0 = 'SELECT COUNT(*), AVG(ended - started) FROM localJobs WHERE ended IS NOT NULL AND ended > '.$oneIntervalAgo.';';
```
mattabullock: Use a better variable name.

bin/BedrockWorkerManager.php
```diff
@@ -458,6 +506,108 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
...
+    $temp0 = $localDB->read($q0);
```
mattabullock: Use a better variable name.

bin/BedrockWorkerManager.php
```diff
@@ -458,6 +506,108 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
...
+    $lastIntervalCount = $temp0[0];
+    $lastIntervalAverage = floatval($temp0[1]);
+
+    $q1 = 'SELECT COUNT(*), AVG(ended - started) FROM localJobs WHERE ended IS NOT NULL AND ended > '.$twoIntervalsAgo.' AND ended < '.$oneIntervalAgo.';';
```
mattabullock: Use a better variable name.

bin/BedrockWorkerManager.php
```diff
@@ -458,6 +506,108 @@ function getNumberOfJobsToQueue(LocalDB $localDB, int $target, int $maxSafeTime,
...
+    $temp1 = $localDB->read($q1);
```
mattabullock: Use a better variable name.

bin/BedrockWorkerManager.php
tylerkaraszewski: > if you made changes here they would perpetuate beyond this function call

I do this on purpose. I want to update the global `$target` and `$lastRun`.

bin/BedrockWorkerManager.php
tylerkaraszewski: I think this actually makes less difference than I thought it did, I'll try it again with the original behavior.

tylerkaraszewski:
@mattabullock - updated.

bin/BedrockWorkerManager.php
mattabullock: Yeah it won't make much difference on local since you're just hitting your loopback, so probably not a lot of network latency. It was making a bigger difference on the servers (we noticed a marked improvement when we made this change originally).

PR #103, opened 2018-10-15
Add commitCount to all updateJob and createJob data

iwiznia:
Needed to fix this specific instance  https://github.com/Expensify/Expensify/issues/88302#issuecomment-429371913

Hold on https://github.com/Expensify/Bedrock/pull/514

Tests:
- Create a job without having done any bedrock call in any cluster. Check it created a job without adding any commitCounts.
- Create a job having done a call in only one cluster. Check it created the job and commitCounts contains only the count for that one cluster.
- Create a job having done a call in more than one cluster. Check it created the job and commitCounts contains all the counts for the clusters.
- Run BWM on the jobs created above and check that they correctly pass the commitCount (note: the webrock commitCount is updated again [here](https://github.com/Expensify/Bedrock-PHP/blob/1236fe20c4c3b82cbb50a56d67d918733edd765f/bin/BedrockWorkerManager.php#L287), which is correct given that we want the commitCount of the GetJobs to be included, which will always be greater than the commitCount for when the job was queued).

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -263,6 +263,14 @@ class Client implements LoggerAwareInterface
     }
 
     /**
+     * Returns the cluster name.
+     */
+    public function getClusterName(): string
+    {
+        return $this->clusterName;
+    }
+
+    /**
      * Makes a direct call to Bedrock.
      *
      * @param string $method  Request method
@@ -592,7 +600,7 @@ class Client implements LoggerAwareInterface
         // If we received the commitCount, then save it for future requests. This is useful if for some reason we
         // change the bedrock node we are talking to.
         if (isset($responseHeaders["commitCount"])) {
-            $this->commitCount = $responseHeaders["commitCount"];
+            $this->commitCount = (int) $responseHeaders["commitCount"];
         }
 
         return [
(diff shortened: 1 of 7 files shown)


iwiznia:
I need to do a couple more tests for this...

iwiznia:
ok, this is ready.

iwiznia:
Will create the web part when this is merged.

src/Client.php
```diff
@@ -375,6 +398,7 @@ public function call($method, $headers = [], $body = '')
             } catch (BedrockError $e) {
                 // This error happen after sending some data to the server, so we only can retry it if it is an idempotent command
                 $this->markHostAsFailed($hostName);
+                /* @phan-suppress-next-line PhanTypeInvalidDimOffset for some reason phan says idempotent does not exist, but I have the ?? so it should not matter */
```
fnwbr: Did you add this after you updated Phan?

src/Jobs.php
```diff
@@ -197,6 +198,10 @@ public function createJobs(array $jobs): array
                 unset($jobs[$i]['priority']);
             }
         }
+        $commitCounts = Client::getCommitCounts();
+        foreach ($jobs as $i => $job) {
```
fnwbr: NAB: I personally prefer `array_walk` with passing the first param as a reference over the `$jobs[$i]` stuff.

```php
array_walk($jobs, function (&$job) use ($commitCounts) {
    $job['data'] = array_merge($job['data'] ?? [], count($commitCounts) ? ['commitCounts' => $commitCounts] : []);
}
```

src/Client.php
iwiznia: Yes, phan was broken and not running (not sure since when)

pecanoro:
Without waiting for my review... 😅

fnwbr:
Ionatan asked me to 😬

src/Jobs.php
iwiznia: Yeah, I'll take the NAB, I don't like that with that awful reference there and everything.

pecanoro: array_walk is way more difficult to read/understand

iwiznia: lol 
image

pecanoro: Maybe if you try with `array_walk` instead of `array_wakk` you will something 😄

iwiznia: You might think so...
image

PR #106, opened 2018-11-09
Only retry timeout that isn't from a SQLite query

flodnv:
@tylerkaraszewski @iwiznia please review

https://github.com/Expensify/Expensify/issues/92336

We only want to retry time outs that weren't caused by long running queries. The latter will be thrown with a different message.

# Tests

Ensured that ConnectionFailure is caught for `555 Timeout`. Ensured that exception bubbles up for any other `555 Timeout peeking command`, etc.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -632,9 +632,10 @@ class Client implements LoggerAwareInterface
             $this->commitCount = (int) $responseHeaders["commitCount"];
         }
 
-        // We treat a timeout as a ConnectionFailure
-        if (intval($codeLine) === 555) {
-            throw new ConnectionFailure('Timeout waiting for bedrock response');
+        // We treat a non-sqlite-query timeout as a ConnectionFailure.
+        // We don't want to retry timed out queries as this will just overwhelm the servers.
+        if ($codeLine === '555 Timeout') {
+            throw new ConnectionFailure('Internal Bedrock command timeout (555 Timeout)');
         }
 
         return [


src/Client.php
```diff
@@ -632,9 +632,10 @@ private function receiveResponse()
             $this->commitCount = (int) $responseHeaders["commitCount"];
         }
 
-        // We treat a timeout as a ConnectionFailure
-        if (intval($codeLine) === 555) {
-            throw new ConnectionFailure('Timeout waiting for bedrock response');
+        // We treat a non-sqlite-query timeout as a ConnectionFailure.
+        // We don't want to retry timed out queries as this will just overwhelm the servers.
+        if ($codeLine === '555 Timeout') {
+            throw new ConnectionFailure('Internal Bedrock command timeout (555 Timeout)');
```
iwiznia: `Internal Bedrock command` sounds like a fancy way of saying `query timeout`... How about:  `Bedrock query timeout (555 Timeout)`?

tylerkaraszewski:
I think I can get on board with this, as it basically treats long command queues as transient problems but slow queries as permanent problems. This might not *always* be the case, but it's probably usually the case. Hopefully the long command queue issue goes away on the new servers, anyway.

src/Client.php
flodnv: This is NOT a query timeout 😅

iwiznia: 🤦‍♂️ :ohnothing:

PR #111, opened 2018-11-23
Update bedrock cache

iwiznia:
Reverts Expensify/Bedrock-PHP#108 see https://github.com/Expensify/PHP-Libs/pull/175 on why.

It also modifies how the cache works a bit and adds some useful methods to it.

/cc @expensifyrandomdrake since you have a PR using the cache. Should not be much of a difference to you .

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -19,7 +19,7 @@ class Cache extends Plugin
      * @param string $name    Name pattern (using LIKE syntax) to read.
      * @param string $version (optional) Specific version identifier (ie, a timestamp, counter, name, etc), defaults to the latest
      *
-     * @return mixed Whatever what saved in the cache
+     * @return mixed Whatever was saved in the cache
      *
      * @throws DoesNotExist
      */
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -20,6 +20,7 @@ class Cache extends Plugin
      * @param string $version (optional) Specific version identifier (ie, a timestamp, counter, name, etc), defaults to the latest
      *
      * @return mixed Whatever what saved in the cache
+     *
      * @throws DoesNotExist
      */
     public function read($name, $version = null)
@@ -39,9 +40,10 @@ class Cache extends Plugin
     /**
      * Gets data from a cache, if it is not present, it computes it by calling $computeFunction and saves the result in the cache.
      *
-     * @param string $name
+     * @param string      $name
      * @param null|string $version
-     * @param callable $computeFunction
+     * @param callable    $computeFunction
+     *
      * @return array
      */
     public function get(string $name, ?string $version, callable $computeFunction)
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -110,7 +110,7 @@ class Cache extends Plugin
      *
      * @return mixed|null
      */
-    public function call(string $method, array $headers, $body = '')
+    private function call(string $method, array $headers, $body = '')
     {
         // Both writing to and reading from the cache are always idempotent operations
         $headers['idempotent'] = true;
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -34,7 +34,7 @@ class Cache extends Plugin
         if ($response['code'] === 404) {
             throw new DoesNotExist('The cache entry could not be found', 666);
         }
-        return is_array($response['body']) ? $response['body'] : json_decode($response['body']);
+        return $response['body'];
     }
 
     /**
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -34,7 +34,7 @@ class Cache extends Plugin
         if ($response['code'] === 404) {
             throw new DoesNotExist('The cache entry could not be found', 666);
         }
-        return json_decode($response['body']);
+        return is_array($response['body']) ? $response['body'] : json_decode($response['body']);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Cache.php
```diff
@@ -11,21 +12,16 @@
...
-
     /**
      * Reads a named value from the cache.  Can optionally request a specific
      * version of that value, if available.
      *
      * @param string $name    Name pattern (using LIKE syntax) to read.
      * @param string $version (optional) Specific version identifier (ie, a timestamp, counter, name, etc), defaults to the latest
      *
-     * @return array Containing "name" (the name matched), "rawBody" (unparsed), and "body" (JSON parsed)
+     * @return mixed Whatever what saved in the cache
```
chiragsalian: NAB: Did you mean to say -> "Whatever **was** saved in the cache"

src/Cache.php
```diff
@@ -34,7 +30,49 @@ public function read($name, $version = null)
...
+    /**
+     * Gets data from a cache, if it is not present, it computes it by calling $computeFunction and saves the result in the cache.
+     *
+     * @param string      $name
+     * @param null|string $version
+     * @param callable    $computeFunction
+     *
+     * @return array
+     */
+    public function get(string $name, ?string $version, callable $computeFunction)
```
chiragsalian: Curious about the purpose of this new function. I don't see it being used anywhere.

src/Cache.php
```diff
@@ -65,7 +101,7 @@ public function write($name, $value, $version = null)
             $headers["name"] = "$name/";
         }
 
-        return $this->call("WriteCache", $headers, $value);
+        $this->call("WriteCache", $headers, json_encode($value));
```
chiragsalian: I see you added json_encode here but you don't decode in the read. Shouldn't it be decoded otherwise `ReadCache` would return always return a string right and then we wouldn't get anything for `$response['code']` and `$response['body']`, right?

src/Cache.php
iwiznia: It adds to the library a pretty common cache pattern: get me this from the cache, if it is not get it from the non-cache and save it. 
It is used in my PR for example [here](https://github.com/Expensify/Web-Expensify/pull/23113/files#diff-c5436be9641d6d6f80662595b52a5d1eR24), you can see that all that logic was in each place that needed this behavior, so I abstracted it here.

src/Cache.php
iwiznia: I had it in the beginning because I thought the same as you, but then realized/remembered that `Client` already `json_decodes` it [here](https://github.com/Expensify/Bedrock-PHP/blob/04492925322ae53f697fd4e13268ec0c9e1bdf38/src/Client.php#L682).

src/Cache.php
```diff
@@ -34,7 +30,49 @@ public function read($name, $version = null)
...
+    /**
+     * Reads from the cache, but if it does not find the entry, it returns the passed default.
+     *
+     * @param string $name
+     * @param mixed  $default
+     * @param string $version
+     *
+     * @return mixed
+     */
+    public function readWithDefault($name, $default, $version = null)
```
fnwbr: Can't we just use `$cache->read($name) ?? $default` instead of this method?

src/Cache.php
fnwbr: 👍 

Might also make `$computeFunction` an optional parameter, and then we could deprecate the `read` method altogether.

src/Exceptions/Cache/NotFound.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Expensify\Bedrock\Exceptions\Jobs;
```
fnwbr: Wrong namespace

src/Exceptions/Cache/NotFound.php
```diff
@@ -0,0 +1,12 @@
...
+
+use Expensify\Bedrock\Exceptions\BedrockError;
+
+/**
+ * Thrown when bedrock can't find the given cache entry.
+ */
+class DoesNotExist extends BedrockError
```
fnwbr: Class name does not match file name

src/Cache.php
iwiznia: No, because `read` throws if the entry is not present.

src/Cache.php
iwiznia: Ummm we can do that, do you think that's better? I am unsure.

src/Exceptions/Cache/NotFound.php
iwiznia: lol

src/Exceptions/Cache/NotFound.php
iwiznia: 🤦‍♂️

iwiznia:
Updated

src/Cache.php
```diff
@@ -3,6 +3,7 @@
 namespace Expensify\Bedrock;
 
 use Expensify\Bedrock\Exceptions\ConnectionFailure;
+use Expensify\Bedrock\Exceptions\Jobs\DoesNotExist;
```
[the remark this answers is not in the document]
fnwbr: Nope

src/Cache.php
fnwbr: Not feeling passionate either way... just `read` and `get` seem interchangeable and weird to be on the same class. NAB

PR #112, opened 2018-11-30
Check for 1 and true in BedrockWorkerManager

qxh5696:
Change the if condition to confirm if a bedrock job is mock job by checking if the `mockRequest` parameter is `true` instead of `1`. 

$ https://github.com/Expensify/Expensify/issues/85244

# Test
1. Uncomment this [line](https://github.com/Expensify/Web-Expensify/blob/dc63631bde51cd4517b6dec8b0137767cd191089/www-dev-config.php#L4) in Web-Expensify
2. Start the bedrock worker manager and tail the logs for the remainder of the steps
3. Trigger a job through `Web-Expensify` (I did a ReportExporter job through IS)
4. Monitor the logs to verify that we created the job with `'mockRequest': 1`
image
5. After the job is completed, look for this log line to verify that we skipped [this section](https://github.com/Expensify/Bedrock-PHP/blob/master/bin/BedrockWorkerManager.php#L313) for the mocked job
image

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -308,7 +308,7 @@ try {
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== true) {
+                                if (!$worker->getParam("mockRequest")) {
                                     // Run the worker.  If it completes successfully, finish the job.
                                     $worker->run();
 
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -308,7 +308,7 @@ try {
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== '1' && $worker->getParam("mockRequest") !== 'true') {
+                                if ($worker->getParam("mockRequest") !== 'true') {
                                     // Run the worker.  If it completes successfully, finish the job.
                                     $worker->run();
 
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -308,7 +308,7 @@ try {
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== '1') {
+                                if ($worker->getParam("mockRequest") !== '1' && $worker->getParam("mockRequest") !== 'true') {
                                     // Run the worker.  If it completes successfully, finish the job.
                                     $worker->run();
 
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -308,7 +308,7 @@ try {
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== 'true') {
+                                if ($worker->getParam("mockRequest") !== true) {
                                     // Run the worker.  If it completes successfully, finish the job.
                                     $worker->run();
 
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -308,7 +308,7 @@
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== '1') {
+                                if ($worker->getParam("mockRequest") !== '1' && $worker->getParam("mockRequest") !== 'true') {
```
qxh5696: Is it safe to omit the check for `1`? I was assuming that we still wanted to keep it to be safe.

iwiznia: Yes, it is useless, remove the 1 check.

iwiznia:
Please add tests to the description.

bin/BedrockWorkerManager.php
```diff
@@ -308,7 +308,7 @@
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== '1') {
+                                if ($worker->getParam("mockRequest") !== 'true') {
```
yuwenmemon: If we look at the job here:
screen shot 2018-11-30 at 10 48 38 am

We can see that `true` is being stored as its boolean value `true` as opposed to the string `"true"`. We should be careful of this, correct?
```
yuwen@MacBook-Pro-99: Bedrock-PHP$ ../script/psysh.sh 
Psy Shell v0.8.18 (PHP 7.2.7-1+ubuntu14.04.1+deb.sury.org+1 — cli) by Justin Hileman
New version is available (current: v0.8.18, latest: v0.9.9)
>>> 'true' === true
=> false
```

yuwenmemon: Yeah or we could just use the falsey/truthy behavior of PHP

qxh5696: I'm running tests to confirm what we should be looking for, but I also think it should be `true` and not `"true"`.

qxh5696:
Updated:
- PR Description includes the correct test steps

bin/BedrockWorkerManager.php
```diff
@@ -308,7 +308,7 @@
 
                             // Open the DB connection after the fork in the child process.
                             try {
-                                if ($worker->getParam("mockRequest") !== '1') {
+                                if ($worker->getParam("mockRequest") !== true) {
```
flodnv: >Is it safe to omit the check for 1? I was assuming that we still wanted to keep it to be safe.

I am with Qadir here, I would've left it. 

>Yes, it is useless, remove the 1 check.

@iwiznia until we make a regression go the other way and `'1'` is back again

>These are the cases when I think that not using strict checking could be useful...

I totally agree. Given the above, I think we should change this check to not strict and we will be covering all possible regressions. There's no value in having a strict check here, but there's value in NOT having it. I'm strongly of the opinion to change this to a non strict check. If you are not, can you make a case for it?

iwiznia: > @iwiznia until we make a regression go the other way and '1' is back again

Yes, every piece of code works till someone introduces a bug...

Anyway, I am fine with making this check a truthy check, but there's many places in which this same case could be made and our style guide says to always use strict checks. I am all up for reviewing that guideline if we want to skip it here (and in other places).

flodnv: Ok, let's make it a truthy check, reality is this is impossible to introduce new bugs like this.

We can revisit the style guide as well

qxh5696:
Updated: 
- Check for `mockRequest` job parameter is now a truthy™️  check.

iwiznia:
Looks good, but you need to create a new version of the lib, using `package.sh`

qxh5696:
Updated the build version

PR #113, opened 2019-01-04
Check $_server for mock request header

bondydaa:
When we moved over to FPM the version we are on doesn't include `getallheaders` so the call errors with an unknown function error.

Discovered apart of https://github.com/Expensify/Expensify/issues/95795 and discussion in #infra https://expensify.slack.com/archives/C094TGUTZ/p1546619668475300

I'm not 100% on changing this since we safe guard with `function_exists` and FPM PHP 7.3 does have `getallheaders`

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -173,7 +173,7 @@ class Client implements LoggerAwareInterface
             $this->mockRequests = $config['mockRequests'];
         } elseif (isset($_SERVER['HTTP_X_MOCK_REQUEST'])) {
             // otherwise check the http headers and set it
-            $this->mockRequests = (bool) $_SERVER['HTTP_X_MOCK_REQUEST'];
+            $this->mockRequests = isset($_SERVER['HTTP_X_MOCK_REQUEST']);
         }
 
         // Make sure we have at least one host configured
--- a/src/Client.php
+++ b/src/Client.php
@@ -173,7 +173,7 @@ class Client implements LoggerAwareInterface
             $this->mockRequests = $config['mockRequests'];
         } else {
             // otherwise check the http headers
-            $this->mockRequests = isset($_SERVER['HTTP_X-Mock-Request']);
+            $this->mockRequests = isset($_SERVER['HTTP_X_MOCK_REQUEST']);
         }
 
         // Make sure we have at least one host configured
--- a/src/Client.php
+++ b/src/Client.php
@@ -171,7 +171,7 @@ class Client implements LoggerAwareInterface
         // If the caller explicitly set `mockRequests`, use that value.
         if (isset($config['mockRequests'])) {
             $this->mockRequests = $config['mockRequests'];
-        } elseif (isset($_SERVER['HTTP_X_MOCK_REQUEST']))_ {
+        } elseif (isset($_SERVER['HTTP_X_MOCK_REQUEST'])) {
             // otherwise check the http headers and set it
             $this->mockRequests = $_SERVER['HTTP_X_MOCK_REQUEST'];
         }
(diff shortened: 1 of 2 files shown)


bondydaa:
working on that still 😬

src/Client.php
```diff
@@ -171,12 +171,9 @@ private function __construct(array $config = [])
...
             $this->mockRequests = $config['mockRequests'];
-        } elseif (function_exists('getallheaders')) {
-            // otherwise pull from the request headers.
-            $requestHeaders = getallheaders();
-            if (isset($requestHeaders['X-Mock-Request'])) {
-                $this->mockRequests = true;
-            }
+        } elseif (isset($_SERVER['HTTP_X_MOCK_REQUEST'])) {
+            // otherwise check the http headers and set it
+            $this->mockRequests = $_SERVER['HTTP_X_MOCK_REQUEST'];
```
coleaeason: this was a bool before but now you're setting it to a string, is that on purpose?

bondydaa: no good point, I assumed it would be a bool.

bondydaa:
I believe I tested this properly but not 100% sure.

I installed this locally by editing my `web-e` composer.json file to have:

```
"expensify/bedrock-php": "dev-bondy-no-getallheaders#38c68134e9a4a0b906c5c0e0e6a372c857114747",
```

Then installed via `composer update expensify/bedrock-php`

Whenever we call `Auth::call` [here](https://github.com/Expensify/Web-Expensify/blob/e62ae87b1fdb3d1b885a12d92d1266322448a0c2/lib/Auth.php#L451), we get the `Client::instance` which should be calling this method.

Here are my logs for creating a new transaction first instance of `CreateTransaction` has no `mockRequest` (expected)

```
2019-01-05T00:31:18.241560+00:00 expensidev ool www: BySxa3 /api.php dbondy@expensifail.com !web! ?api? [info] Bedrock\Client - Starting a request ~~ command: 'CreateTransaction' clusterName: 'auth' headers: '[created: '2019-01-04' merchant: 'butts-r-us-again' amount: '22200' currency: 'USD' category: 'Uncategorized' comment: '{"comment":""}' mcc: '' billable: 'false' reimbursable: 'true' tag: '' receiptID: '' reportID: '157' commitCount: '50955' requestID: 'BySxa3' lastIP: '10.2.2.1' writeConsistency: 'ASYNC' priority: '500' timeout: '290000']'
```

Subsequent calls in logs for `CreateTransaction` has `mockRequest`:
```
2019-01-05T00:31:21.502221+00:00 expensidev ool www: 3Dj47r /api.php dbondy@expensifail.com !web! ?api? [info] Bedrock\Client - Starting a request ~~ command: 'CreateTransaction' clusterName: 'auth' headers: '[created: '2019-01-04' merchant: 'butts-r-us-again' amount: '22200' currency: 'USD' category: 'Uncategorized' comment: '{"comment":""}' mcc: '' billable: 'false' reimbursable: 'true' tag: '' receiptID: '' reportID: '157' commitCount: '50971' requestID: '3Dj47r' lastIP: '10.2.2.1' writeConsistency: 'ASYNC' mockRequest: '1' priority: '500' timeout: '290000']'
```

src/Client.php
coleaeason:
```
>>> (bool) 'true';
=> true
>>> (bool) 'false';
=> true
>>> (bool) 'literally-any-string';
=> true
>>> (bool) '';
=> false
```
you sure that works?

bondydaa: 🤦‍♂️

bondydaa:
pushed again, also found a small issue in the composer.json that I fixed

PR #114, opened 2019-01-09
Create versionWatchFile if it doesn't exist

mattabullock:
If you add the versionWatchFile flag but the file doesn't exist, create it.

Part of https://github.com/Expensify/Expensify/issues/95570

# Tests
1. Make sure version watch file doesn't exist
2. Run BWM with `--versionWatchFile=<file>` flag
3. Confirm that BWM runs

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -107,8 +107,11 @@ if ($enableLoadHandler) {
     $localDB->write($query);
 }
 
-// If --versionWatch is enabled, begin watching a version file for changes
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -107,8 +107,11 @@
     $localDB->write($query);
 }
 
-// If --versionWatch is enabled, begin watching a version file for changes
+// If --versionWatch is enabled, begin watching a version file for changes. If the file doesn't exist, create it.
 $versionWatchFile = @$options['versionWatchFile'];
```
flodnv: NAB because you didn't touch this, but we should NEVER be using `@`. `??` is what should be used here.

PR #115, opened 2019-01-15
Ignore 115 errors

tylerkaraszewski:
`115 operation in progress` isn't a fatal error condition, so let's not treat it as such.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -487,7 +487,7 @@ class Client implements LoggerAwareInterface
             socket_set_option($this->socket, SOL_SOCKET, SO_RCVTIMEO, ['sec' => $this->readTimeout, 'usec' => 0]);
             @socket_connect($this->socket, $host, $port);
             $socketErrorCode = socket_last_error($this->socket);
-            if ($socketErrorCode) {
+            if ($socketErrorCode && $socketErrorCode !== 115) {
                 $socketError = socket_strerror($socketErrorCode);
                 throw new ConnectionFailure("Could not connect to Bedrock host $host:$port. Error: $socketErrorCode $socketError");
             }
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -487,7 +487,9 @@ private function sendRawRequest(string $host, int $port, string $rawRequest)
             socket_set_option($this->socket, SOL_SOCKET, SO_RCVTIMEO, ['sec' => $this->readTimeout, 'usec' => 0]);
             @socket_connect($this->socket, $host, $port);
             $socketErrorCode = socket_last_error($this->socket);
-            if ($socketErrorCode) {
+            if ($socketErrorCode === 115) {
+                $this->logger->info('Bedrock\Client - socket_connect returned error 115, continuing.');
```
iwiznia: I don't get this, just continue as if it succeeded? Don't you need to throw a `ConnectionFailure` so we retry in another server?

righdforsa: Correct, we're just going to continue as if it succeeded.

flodnv:
FYI, this results in a lot of PHP Warnings: https://github.com/Expensify/Expensify/issues/96847

iwiznia:
> FYI, this results in a lot of PHP Warnings: Expensify/Expensify#96847

Of course it did, this change makes no sense. I should've rejected instead of commented...

I see 0 tests here and it was CPed to prod...

quinthar:
As discussed in https://github.com/Expensify/Expensify/issues/96847#issuecomment-455242389 this is known and intentional.  I think we can close this.

iwiznia:
We can solve the warning, plus if you close that issue, botify will reopen it.

PR #116, opened 2019-01-16
Add retryAfter to queueJob function

mattabullock:
We need to add the retryAfter param to queueJob for an upcoming change on web. It already exists in CreateJob so we just need to accept the param and pass it through.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "1.5.6",
+    "version": "1.5.7",
     "authors": [
         {
             "name": "Expensify",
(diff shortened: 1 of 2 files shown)


mattabullock:
I'm more than happy to just switch all the calls to CreateJob instead of QueueJob. Is there a historical reason for silencing errors?

iwiznia:
No idea, my guess is that we call this on jobs that if don't get created, we don't care.

mattabullock:
Gonna close this and just migrate the QueueJob calls to CreateJob.

mattabullock:
@iwiznia we're overruling you, don't wanna open up spots in the code where we can exit early and create more errors for jobs we previously didn't care about.

iwiznia:
I am not sure I follow what you said, but go ahead.

PR #123, opened 2019-01-28
Remove logParam from getInstance method

iwiznia:
Fixes Expensify/Expensify#96880

Check that commitCount is correctly preserved for webrock in BWM.

The getInstance method uses the config passed and the default config to build a new instance of the Client or to return an existing one that has been built before that has the same configuration. When we added the `logParam` config, it broke this system in BWM, because we first configure it with logParam `null` and then we re-configure it with logParam as the email of the user that owns the job being run, that made it so it returned a new instance of the Client for each job, one that did not have the commitCount set correctly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "1.6.2",
+    "version": "1.6.4",
     "authors": [
         {
             "name": "Expensify",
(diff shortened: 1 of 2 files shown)


flodnv:
https://github.com/Expensify/Expensify/issues/96880#issuecomment-457698032

>Will add more context on monday when I submit the PR for review.

Can you add a tad more context please? 😄

iwiznia:
Added it to the description

PR #124, opened 2019-01-28
Fix timing log

iwiznia:
@coleaeason please review

Related to Expensify/Expensify#70743 (comment)

Tests:
```
2019-01-24T13:27:34.834042+00:00 expensidev bedrock: xGChxM (BedrockCommand.cpp:286) finalizeTimingInfo [worker3] [info] command 'SetNameValuePair' timing info (ms): 5 (1), 0 (0), 0, 0, 0, 0, 5, 0, 0. Upstream: 0, 0, 0, 0.
2019-01-24T13:27:36.983074+00:00 expensidev php-cgi: xGChxM /caca.php ionatan@expensify.com !web! ?api? [info] Bedrock\Client - Request finished ~~ host: 'auth' command: 'SetNameValuePair' jsonCode: '200 OK' duration: '11' net: '5.619' wait: '5.381' proc: '0' commitCount: '14927'
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "1.6.4",
+    "version": "1.6.5",
     "authors": [
         {
             "name": "Expensify",
--- a/src/Client.php
+++ b/src/Client.php
@@ -457,9 +457,9 @@ class Client implements LoggerAwareInterface
         }
 
         // Log how long this particular call took
-        $processingTime = isset($response['headers']['processTime']) ? $response['headers']['processTime'] : 0;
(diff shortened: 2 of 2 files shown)


iwiznia:
> What unit is processing/server/clientTime?

Times coming from bedrock are in useconds, time computed in PHP via the `microtime(true)` is in seconds. We convert all to miliseconds.

quinthar:
>  time computed in PHP via the microtime(true) is in seconds. We convert all to miliseconds.
Ah, well if you did `microtime()` then it'd return in microseconds -- just like the others -- and you could divide by 1000 and be consistent.

iwiznia:
> Ah, well if you did microtime() then it'd return in microseconds 

No, it [would not](http://php.net/manual/en/function.microtime.php).
image

quinthar:
Whoops, as always, you are right.  Thanks for the correction!

PR #126, opened 2019-02-08
Respect connection close coming from the server

iwiznia:
Fixes issues like what we see in [this request](https://www.expensify.com/_devportal/tools/logSearch/#query=request_id:(%2202Mdtd%22)+AND+timestamp:%5B2019-01-23T14:41:11.607Z+TO+2019-01-23T16:41:11.607Z%5D&index=logs_expensify-001259)

Tests:
- Make a bedrock request during server shutdown
- Check Connection:close is returned by the server
- Check client disconnects from the server and establishes a new connection

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -383,7 +383,7 @@ class Client implements LoggerAwareInterface
         // If we passed a preferred host and we already had a connected socket, but to a different host and the preferred
         // host is not blacklisted (the preferred host is returned first in the possible hosts array only when it's not blacklisted)
         // then we close the socket in order to connect to the preferred one.
-        $closeSocketAfterRequest = array_key_exists('Connection', $headers) ? $headers['Connection'] : false;
+        $closeSocketAfterRequest = array_key_exists('Connection', $headers) ? $headers['Connection'] === 'close' : false;
 
         if ($preferredHost && $this->socket && key($hostConfigs) !== $this->lastHost) {
             @socket_close($this->socket);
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -383,7 +383,7 @@ public function call($method, $headers = [], $body = '')
         // If we passed a preferred host and we already had a connected socket, but to a different host and the preferred
         // host is not blacklisted (the preferred host is returned first in the possible hosts array only when it's not blacklisted)
         // then we close the socket in order to connect to the preferred one.
-        $closeSocketAfterRequest = array_key_exists('Connection', $headers) ? $headers['Connection'] : false;
+        $closeSocketAfterRequest = array_key_exists('Connection', $headers) ? $headers['Connection'] === 'close' : false;
```
tylerkaraszewski: This is here to catch `Connection: forget`, right?

We receive `close` as a response, but we send `forget` when we want to close the connection.

I'm not sure we need this change, or if we do, do we want to check for either `close` or `forget`?

iwiznia: I thought the 2 were different... 
forget: I don't need to receive a response from this command, server can process it later
close: Close this connection after sending the request

So, this was closing the connection after any request with `Connection:forget` which is not correct....
Am I misunderstanding this?

tylerkaraszewski: Never mind, I'm interpreting this incorrectly.

PR #127, opened 2019-02-15
Remove AIMD1 in favor of AIMD2

flodnv:
@mattabullock or @tylerkaraszewski please review


flodnv:
Should we also kill `enableLoadHandler` ?

mattabullock:
>Should we also kill `enableLoadHandler` ?
I'd say leave it, nice to have a kill switch that doesn't require two deploys.

flodnv:
You can merge and then this https://github.com/Expensify/Web-Expensify/pull/24091

PR #128, opened 2019-02-22
Silence built-in warning, we'll handle it

tylerkaraszewski:
@iwiznia 

Fixes: https://github.com/Expensify/Expensify/issues/96913

All this does is silence the warning here. We already handle the error case here with our retry logic exactly how we're supposed to, but we don't keep the warning from logging. Sometimes after `connect` returns `115`, we get this error. The retry should handle it correctly.

I'll update the version number if you agree with this change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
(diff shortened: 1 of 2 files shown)


tylerkaraszewski:
@iwiznia updated.

iwiznia:
Don't forget to update the lib in web and secure.

PR #129, opened 2019-03-04
Catch BWM crashing when jobID doesn't exist

mattabullock:
We already attempt to catch if the job was retried elsewhere, so that would mean the jobID doesn't exist. The error name must've changed or something...

Part of https://github.com/Expensify/Expensify/issues/95570

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -352,7 +352,7 @@ try {
                                 ]);
                                 try {
                                     $jobs->retryJob((int) $job['jobID'], $e->getDelay(), $worker->getData(), $e->getName(), $e->getNextRun());
-                                } catch (IllegalAction $e) {
+                                } catch (IllegalAction|DoesNotExist $e) {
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -339,7 +339,7 @@
                                 ]);
                                 try {
                                     $jobs->retryJob((int) $job['jobID'], $e->getDelay(), $worker->getData(), $e->getName(), $e->getNextRun());
-                                } catch (IllegalAction $e) {
+                                } catch (IllegalAction | DoesNotExist $e) {
```
iwiznia: If this makes sense here, it should also make sense in the other places like line 358 and 326, right?

mattabullock: We already catch it at 326, but I can add it to 358 too.

iwiznia: Oh right, did not see the other catch. I think we should add it to 358.

mattabullock: I'm trying to but github won't let me assign you the pr or edit the title.

PR #133, opened 2019-05-07
Retry in all hosts again if all hosts failed

iwiznia:
Attempts to fix the issue https://github.com/Expensify/Expensify/issues/69763

Holding to address any comments and create the new version.

Tests:
- Stop the auth server and make a request. Check it tries all possible hosts twice, the fails.
- Stop the auth server and make a request. Check it tries all possible hosts once, then start the server back up and check it finishes the request correctly.
- Modify code so that instead of ConnectionError it throws BedrockError, stop the auth server and make an idempotent request, check it tries to use all the hosts twice then fails.
- Modify code so that instead of ConnectionError it throws BedrockError, stop the auth server and make an idempotent request, check it tries to use all the hosts once, then start the server back and check it finishes the request correctly.
- Modify code so that instead of ConnectionError it throws BedrockError, stop the auth server and make a non idempotent request, check it throws immediately after fail.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -392,7 +392,9 @@ class Client implements LoggerAwareInterface
         }
 
         $hostName = null;
+        $retriedAllHosts = false;
         while (!$response && count($hostConfigs)) {
+            $exception = null;
             reset($hostConfigs);
             $numRetriesLeft = count($hostConfigs) - 1;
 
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -581,7 +610,7 @@ private function getPossibleHosts(string $preferredHost = null)
         shuffle($failoverHostNames);
         $mainHostName = array_rand($this->mainHostConfigs);
         $preferredHost = array_key_exists((string) $preferredHost, $cachedHostConfigs) ? [$preferredHost] : [];
-        $hostNames = array_unique(array_merge($preferredHost, [$mainHostName], $failoverHostNames));
+        $hostNames = array_filter(array_unique(array_merge($preferredHost, [$mainHostName], $failoverHostNames)));
```
flodnv: Why did you need to add this? 😕

src/Client.php
```diff
@@ -425,16 +427,21 @@ public function call($method, $headers = [], $body = '')
...
                 $this->markHostAsFailed($hostName);
                 /* @phan-suppress-next-line PhanTypeInvalidDimOffset for some reason phan says idempotent does not exist, but I have the ?? so it should not matter */
-                if ($numRetriesLeft && ($headers['idempotent'] ?? false)) {
-                    $this->logger->info('Bedrock\Client - Failed to send the whole request or to receive it; retrying because command is idempotent', ['host' => $hostName, 'message' => $e->getMessage(), 'retriesLeft' => $numRetriesLeft, 'exception' => $e]);
+                if ($headers['idempotent'] ?? false) {
+                    if ($numRetriesLeft) {
+                        $this->logger->info('Bedrock\Client - Failed to send the whole request or to receive it; retrying because command is idempotent', ['host' => $hostName, 'message' => $e->getMessage(), 'retriesLeft' => $numRetriesLeft, 'exception' => $e]);
+                    } else {
+                        $this->logger->error('Bedrock\Client - Failed to send the whole request or to receive it; not retrying', ['host' => $hostName, 'message' => $e->getMessage(), 'exception' => $e]);
```
flodnv: specify `not retrying because we are out of retries` ?

src/Client.php
```diff
@@ -443,6 +450,23 @@ public function call($method, $headers = [], $body = '')
...
+
+            // All non blacklisted hosts failed, this could be because we are in the middle of a cluster version flip.
+            // ie: We have version 1 and version 2, we've installed version 2 in less than half the cluster, a previous
+            // request already marked all servers in version 2 as failed (since 1 is the current version) in this request
+            // we have installed version 2 in one more server, making it the current version. So now all the servers that
+            // were marked as failed in the previous request are the ones serving requests and all the ones that were good
+            // before are now in the old version and not serving requests. So to cover this, we retry in all servers
+            // once hoping it will find a server that works.
+            if ($exception) {
+                if (!$retriedAllHosts) {
```
flodnv: NABish you can make this a fail fast:
```
if $exception
   if $retriedAllHosts
       throw
   ...
```

src/Client.php
iwiznia: Because I realized that `preferredHost` could be null (see the change in the type annotation) and without it we were leaving a `null` in the hostNames... I think it was not causing any problem, but this is more correct.

iwiznia:
Updated and created the version

PR #136, opened 2019-05-17
Adding a function to call our bedrock endpoint of RequeueJobs

chiragsalian:
@pecanoro, can you review this. 

We have a RequeueJobs method in Bedrock. This is just an intermediate function so that web-E can make use of that bedrock endpoint.

~On HOLD till this Bedrock [PR](https://github.com/Expensify/Bedrock/pull/643) gets merged.~

Helps with this [PR](https://github.com/Expensify/Web-Expensify/pull/24807) 

# Tests / QA
None

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/Bedrock-PHP",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "1.7.6",
+    "version": "1.7.7",
     "authors": [
         {
             "name": "Expensify",
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -370,6 +370,21 @@ class Jobs extends Plugin
     }
 
     /**
+     * Requeues a job back to QUEUED state. Updates the job name if provided as well.
+     */
+    public function requeueJobs(array $jobIDs, string $name = ''): array
+    {
+        return $this->call(
+            "RequeueJobs",
+            [
+                "jobIDs" => implode(',',$jobIDs),
+                "name" => $name,
+                "idempotent" => true,
+            ]
+        );
+    }
+
+    /**
      * Query a job's info.
      * Bedrock will return:
      *     - 200 - OK
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -377,7 +377,7 @@ class Jobs extends Plugin
         return $this->call(
             "RequeueJobs",
             [
-                "jobIDs" => implode(',',$jobIDs),
+                "jobIDs" => implode(',', $jobIDs),
                 "name" => $name,
                 "idempotent" => true,
             ]


pecanoro:
Why on WIP? Also, why did we new command `RequeueJobs` instead of `RetryJobs`? I am just curious

chiragsalian:
WIP since i have to update version and tag :) 

>  why did we new command RequeueJobs instead of RetryJobs?

Coming from this [discussion](https://github.com/Expensify/Expensify/issues/103633#issuecomment-493415723). RetryJob has limitation where we can only retry a RUNNING job. So if we do not use RequeueJobs we'll have to first call QueryJob to figure out its state and then call either UpdateJob or RetryJob. We've to also update UpdateJob to support updating the job name. 

Instead we decided we'll update RequeueJobs to support updating the job name and then we don't have to worry about the state of the job.

chiragsalian:
Updated version and tag. Its out of WIP now :)

quinthar:
Have we considered just updating the behavior of retry job to do what we want? It does feel suspiciously like the same command, and it would be nice to not have two newly identical commands that have slightly different behavior

chiragsalian:
Hey david, yes we briefly discussed about that [here](https://github.com/Expensify/Expensify/issues/103633#issuecomment-493415723). `retryJob` was specifically created to only retry a job that is currently running. It will not retry a job that is QUEUED. I was a little hesitant to update retryJob to handle QUEUED jobs since it has specific code to throw errors for QUEUED jobs.

Hence my choices were,
- to first call queryJob to get the state of the job and then use the existing updateJob or retryJob depending on the state of the job to update the jobName, or
- to use RequeueJobs which doesn't depend on the state of the job. 

I used the latter since it seemed like the least calls to bedrock. Do you feel like we should do this differently?

quinthar:
Maybe I'm not understanding, but why do you need to requeue a job if it is already queued? What does that actually do if anything?

chiragsalian:
Well we don't requeue the job if its already queued. In our case if the job is queued all we do is update the jobName. 

So here is the scenario, assume we have these 4 assigned jobs,
```
QUEUED      manual/ChatbotUnprocessedInput?queue=c4b&accountID=1
QUEUED      manual/ChatbotUnprocessedInput?queue=firstResponder&accountID=1
RUNQUEUED      manual/ChatbotUnprocessedInput?queue=c4b&accountID=1
RUNQUEUED      manual/ChatbotUnprocessedInput?queue=supportal&accountID=1
```

When the agent clicks Exit we want to unassign these jobs to become,
```
QUEUED      manual/ChatbotUnprocessedInput?queue=c4b
QUEUED      manual/ChatbotUnprocessedInput?queue=firstResponder
QUEUED      manual/ChatbotUnprocessedInput?queue=c4b
QUEUED      manual/ChatbotUnprocessedInput?queue=supportal
```

So our solutions were,
- to first call queryJob to get the state of the job. If the state is queued, we call updateJob and update the name of the job (this would be new bedrock code). If the state is runqueued w call retryJob to update the name of the job and requeue it.
OR
- we just call a single command `RequeueJobs` and the php layer doesn't care of the state of the job and just knows that it'll end up QUEUED.

quinthar:
Got it, that makes sense to me and seems pretty different than retry. Thanks!

pecanoro:
Before merging, how did you test this PR?

chiragsalian:
Bedrock PR has the following automated tests - [here](https://github.com/Expensify/Bedrock/pull/643/files#diff-393fd05b24988b8be0d96c3bc11d038aR1)
and the Web PR will be testing the same in this [PR](https://github.com/Expensify/Web-Expensify/pull/24807)

PR #138, opened 2019-08-05
Allow unparsed Bedrock responses to PHP

tylerkaraszewski:
This change allows for  Bedrock to return non-JSON data to PHP without PHP failing to parse it as JSON. This is being used for this: https://github.com/Expensify/Expensify/issues/103358 to pass a gzipped blob of scoring data back from Bedrock to PHP

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -302,17 +302,16 @@ class Client implements LoggerAwareInterface
     /**
      * Makes a direct call to Bedrock.
      *
-     * @param string $method            Request method
-     * @param array  $headers           Request headers (optional)
-     * @param string $body              Request body (optional)
-     * @param bool   $parseResponseBody Allow passing the body through unparsed and unmodified. (optional)
+     * @param string $method  Request method
+     * @param array  $headers Request headers (optional)
+     * @param string $body    Request body (optional)
      *
      * @return array JSON response, or null on error
      *
      * @throws BedrockError
      * @throws ConnectionFailure
      */
-    public function call($method, $headers = [], $body = '', $parseResponseBody = true)
+    public function call($method, $headers = [], $body = '')
     {
         // Start timing the entire end-to-end
         $timeStart = microtime(true);
@@ -419,7 +418,7 @@ class Client implements LoggerAwareInterface
                 // trying to use might not be in the picked host configs, because getPossibleHosts randomizes them.
                 $port = $this->mainHostConfigs[$hostName]['port'] ?? $this->failoverHostConfigs[$hostName]['port'];
                 $this->sendRawRequest($hostName, $port, $rawRequest);
-                $response = $this->receiveResponse($parseResponseBody);
+                $response = $this->receiveResponse();
             } catch (ConnectionFailure $e) {
                 // The error happened during connection (or before we sent any data, or in a case where we know the
                 // command was never processed) so we can retry it safely.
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -638,11 +639,13 @@ private function getPossibleHosts(?string $preferredHost, bool $resetHosts = fal
...
      * Receives and parses the response.
      *
+     * @param bool $parseResponseBody flag indicating if the body should be parsed or passed unmodified.
+     *
      * @return array Response object including 'code', 'codeLine', 'headers', `size` and 'body'
      *
      * @throws BedrockError
      */
-    private function receiveResponse()
+    private function receiveResponse(bool $parseResponseBody = true)
```
iwiznia: Make this param non optional, given you are always passing it

tylerkaraszewski:
@iwiznia Updated.

iwiznia:
Did you miss my comment above?

tylerkaraszewski:
no? 😕

iwiznia:
lol, then what do you think? I think this should be automatically handled by the headers.

tylerkaraszewski:
Oh, I did miss your comment. I didn't miss your comment on the code, I didn't see the comment above it:

> Can you give me more context on this? What are we going to do with the response after its returned from bedrock?

Pass it unmodified to JS.

> Mainly I ask because it seems like gzip compression should be handled automatically by bedrock, something that I think already exists here.

I don't know what you mean here. In this case, Bedrock will be returning GZipped content, and I don't want the PHP library to decompress it and then parse the result as JSON and then pass that back to the caller, I want PHP to simply pass the compressed blob to the browser without touching it.

I'm doing this because the GZipped content, when unzipped, is several megabytes in size, and PHP does not need to parse that, and (particularly on offshore, etc) we do not need to deliver that content to clients at full size.

What we're delivering is a tree structure of scoring information for Concierge answers. The only place where we need to look at that data is  in the Concierge front-end code.

iwiznia:
Thanks for the context! 
So, before now, all commands to bedrock/auth returned JSON and now we want to support another content type, I think bedrock should return the content-type of the response (defaulted to application/json so all existing commands work the same way) and then add support in `parseRawBody` for the new content type that does not do any conversion.

tylerkaraszewski:
So I Was ready to do what you suggested and was looking at it a little bit, particularly [this answer here](https://stackoverflow.com/a/23600787/9329286) and I wasn't sure what to do. In HTTP, as that answer indicates, the correct way to serve compressed JSON is with:

```
Content-Type: application/json
Content-Encoding: gzip
```

If we did the same, all it indicates to PHP is that it's getting compressed JSON, and should still be able to treat it as JSON.

We could use a custom `Content-Type` header to indicate this, something like `application/json-compressed`, or we can just not care that much and use `application/gzip`.

Thoughts?

tylerkaraszewski:
I'm going to push the change to treat blank and `application/json` as parsable, and anything else as not parsable. I wont increment the version until I get your OK.

tylerkaraszewski:
Ok, as it is now, if we send `application/gzip` from bedrock, we'll pass the body unmodified, which works as intended. If you're happy with this I'll update the package and we can merge.

iwiznia:
> application/gzip from bedrock

Not sure if this is the most "correct" content type, but it really does not matter much.
Anyway, this looks good, create the version.

tylerkaraszewski:
@iwiznia done!

PR #139, opened 2019-08-13
Do not call failJob when finishJob fails on retryAfter jobs

iwiznia:
@pecanoro please review

Did what I proposed [here](https://github.com/Expensify/Expensify/issues/100552#issuecomment-500845895).

# Tests
Queue 2 jobs one with retry after and one without it
Modify bwm in order to fail the  finishJob call
Run bwm, check one of the jobs is in FAILED state and the other one in RUNQUEUED.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use Expensify\Bedrock\Client;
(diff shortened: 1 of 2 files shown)


pecanoro:
Are you sure this is ok? What if `FinishJob` starts failing indefinitely?

iwiznia:
Great point, updated, retested

iwiznia:
I re-made the tag BTW

PR #146, opened 2020-05-20
Allow to pass nextRun in update job

iwiznia:
Hold https://github.com/Expensify/Bedrock/pull/799

Needed for Expensify/Expensify#131401

Tests in the web PR

### DIFF
--- a/src/Jobs.php
+++ b/src/Jobs.php
@@ -271,7 +271,7 @@ class Jobs extends Plugin
      *
      * @return array
      */
-    public function updateJob($jobID, $data, $repeat = null, ?int $priority = null)
+    public function updateJob($jobID, $data, $repeat = null, ?int $priority = null, ?string $nextRun = null)
     {
         $commitCounts = Client::getCommitCounts();
         return $this->call(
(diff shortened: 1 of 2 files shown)


src/Jobs.php
```diff
@@ -271,7 +271,7 @@ public function getJobs(string $name, int $numResults, array $params = []): arra
      *
      * @return array
      */
-    public function updateJob($jobID, $data, $repeat = null, ?int $priority = null)
+    public function updateJob($jobID, $data, $repeat = null, ?int $priority = null, ?string $nextRun = null)
```
alex-mechler, line 274: Need to add the new param to the doc comment

iwiznia: Why? It won't give much more information

alex-mechler: Every other parameter for this already has them, regardless of if they add information, so it feels odd to not have it. Willing to have this a NAB though 🤷‍♂️

iwiznia: Yeah, our stance is add it if it adds more information than the code... I could add a comment about it, but seems simple enough that it does not need one IMO. Happy to add it if you think it is.

iwiznia:
Oh actually we don't need to hold... this will do nothing till bedrock is deployed

PR #147, opened 2020-06-02
Update PHP version to 7.4.3 and update php fixer

tgolen:
Looks like everything is in working order. Gotta see if Travis passes, then we should be good to go here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 php:
-    - '7.2.22'
+  - '7.2.22'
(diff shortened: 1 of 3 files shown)


iwiznia:
FUCK ME! You did not create the version and I merged 🤦

iwiznia:
Sent the PR to bump the version

tgolen:
Ah, I thought about that twice yesterday but never made it happen. Thanks
for the follow up.

PR #149, opened 2020-08-20
Fix jobs to queue metric

justinpersaud:
`bedrockWorkerManager.numberOfJobsToQueue` is using the wrong value for the metric, it's duplicating the target. PR is to fix that.

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -176,7 +176,7 @@ try {
 
         // Check to see if BWM was able to get jobs on the first attempt. If not, it would add a full second each time it failed, skewing the timer numbers.
         if ($isFirstTry) {
-            $stats->timer('bedrockWorkerManager.fullLoop', microtime(true) - $loopStartTime); /* @phan-suppress-current-line PhanTypeMismatchArgument */
+            $stats->timer('bedrockWorkerManager.fullLoop', microtime(true) - $loopStartTime);
         }
 
         // Poll the server until we successfully get a job
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -176,7 +176,7 @@ try {
 
         // Check to see if BWM was able to get jobs on the first attempt. If not, it would add a full second each time it failed, skewing the timer numbers.
         if ($isFirstTry) {
-            $stats->timer("bedrockWorkerManager.fullLoop", microtime(true) - $loopStartTime); /* @phan-suppress-current-line PhanTypeMismatchArgument */
+            $stats->timer('bedrockWorkerManager.fullLoop', microtime(true) - $loopStartTime); /* @phan-suppress-current-line PhanTypeMismatchArgument */
         }
 
         // Poll the server until we successfully get a job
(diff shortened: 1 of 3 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -176,7 +176,7 @@
 
         // Check to see if BWM was able to get jobs on the first attempt. If not, it would add a full second each time it failed, skewing the timer numbers.
         if ($isFirstTry) {
-            $stats->timer("bedrockWorkerManager.fullLoop", microtime(true) - $loopStartTime);
+            $stats->timer("bedrockWorkerManager.fullLoop", microtime(true) - $loopStartTime); /* @phan-suppress-current-line PhanTypeMismatchArgument */
```
iwiznia, line 179: Why the suppress? What was the error? I bet it can be fixed by casting the 2nd param to int, which is what timer expects

bin/BedrockWorkerManager.php
```diff
@@ -163,7 +163,7 @@
                 sleep(1);
             } elseif ($jobsToQueue > $minSafeJobs / 2) {
                 $logger->info('Safe to start a new job, checking for more work', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
-                $stats->timer('bedrockWorkerManager.numberOfJobsToQueue', $target);
+                $stats->timer('bedrockWorkerManager.numberOfJobsToQueue', $jobsToQueue);
```
iwiznia, line 166: lol copy pasta error since years ago 😂

bin/BedrockWorkerManager.php
iwiznia: We don't have the style checker in this repo, can you change the quotes to single quotes? A bit nitpicky but since you need to change this line anyway...

bin/BedrockWorkerManager.php
```diff
@@ -211,6 +211,7 @@
 
         // Found a job
         $loopStartTime = microtime(true);
+        /* @phan-suppress-next-line PhanTypeArraySuspiciousNullable */
```
iwiznia, line 214: Phan, too stupid to realize this can never be null 🤦

iwiznia: Since you added this to 3 places, what about doing something like:
```
$response = $response ?? [];
```
Then we should be able to remove all the suppress you added.
Alternatively, use `??` in all the places we use it, ie: `if (($response['code'] ?? null) == 200)`

iwiznia (changes requested):

bin/BedrockWorkerManager.php
justinpersaud: Oh, this one complains about the type on the interface. It expects a `int` but `microtime` returns a float. In the statsd class, this can take int or float because it rounds the number. I tried changing the interface to accept `number` but that didn't work, is there a better solution?

iwiznia: `StatsInterface` has it as int, so I guess you can update that one and the error will go away

justinpersaud:
updated

iwiznia (approved):
You need to create the version

justinpersaud:
I have the tag here already, for the next part, do we just bump the version in Web-Expensify composer file?

justinpersaud:
I am self merging this because the web-PR was merged.

PR #151, opened 2020-10-27
Lock composer version

alex-mechler:
The travis run is the test 😂

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,11 @@ language: php
 php:
   - '7.4.7'
 
+before_install:
+  - cd ../ && composer self-update 1.6.4 && cd -
+
 install:
   - pecl install ast-1.0.3
-  - cd ../ && composer self-update 1.6.4 && cd -
   - composer install --prefer-dist --no-interaction --dev
 
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,11 +1,10 @@
 language: php
 php:
-  - '7.2.22'
-  - '7.4.3'
+  - '7.4.7'
```
alex-mechler, line 3: Updated this to match the other PHP repos.

bondydaa: yup I think that makes sense with the changes here https://github.com/Expensify/Web-Expensify/pull/28804

cc @iwiznia @flodnv just in case this borks something up

bondydaa (dismissed):

bondydaa:
image

saw the warning but i think that's okay b/c it does show it installed the right version - right?

bondydaa:
oh nvm i see you latest push... will wait for 42 hours til travis finishes again

alex-mechler:
Yeah. Technically our previous version worked, but it still through that error. To prevent people frmo jumping down rabbit holes trying to fix it, i figured it was best to fix now. Catch you in a month when travis finishes

flodnv (approved):

.travis.yml
iwiznia: Oh yeah, I must've forgotten to update this repo!

iwiznia (approved):

PR #154, opened 2020-11-25
Fix running job names containing a '/' after the '?'

flodnv:
For https://github.com/Expensify/Expensify/issues/103058

Jobs named like `www-prod/Worker?email=something/@domain.com` don't run because we log `No worker found, ignoring ~~ jobName: 'www-prod/Worker'`. This fixes that

### Tests

1. Reproduced locally by creating such a job with this name
2. Test again with the fix and see the job worker run
3. Ensure other job workers aren't affected by seeing them run as well

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -243,12 +243,10 @@ try {
                     });
                     $localJobID = $localDB->getLastInsertedRowID();
                 }
-                $parts = explode('/', $job['name']);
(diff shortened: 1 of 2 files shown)


flodnv:
HOLDing on creating the new version til you approve

iwiznia (dismissed):

flodnv:
off hold

iwiznia (approved):

PR #157, opened 2021-03-25
testing if phan is actually working

bondydaa:


bondydaa:
confirmed phan is still throwing as expected on https://travis-ci.com/github/Expensify/Bedrock-PHP/builds/221215986 closing this out.

PR #158, opened 2021-04-02
Log RetryableException message

Beamanator:
## Details

TL;DR: Logs the entire `RetryableException` when caught in `BedrockWorkerManager.php`.

This is useful so we can have more details about the `RetryableException` was thrown.

## Tests

1. Create a bedrock job locally
    1. In the bedrock job's `run` method, throw a `RetryableException` - pass some string into the first parameter of the exception
    2. Example: see how it's being thrown in https://github.com/Expensify/Web-Expensify/pull/30574
2. Run the job, make sure the `RetryableException` gets thrown
3. See the exception data in the logs (search for `"Job could not complete, retrying"`, the exception's data should be shown nearby)

## Fixes
$ https://github.com/Expensify/Expensify/issues/159221

## Notes

Discussion here: https://expensify.slack.com/archives/C03TQ48KC/p1617365746278500

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -349,7 +349,7 @@ try {
                                     'name' => $job['name'],
                                     'id' => $job['jobID'],
                                     'extraParams' => $extraParams,
-                                    'message' => $e->getMessage(),
+                                    'exception' => $e,
                                 ]);
                                 try {
                                     $jobs->retryJob((int) $job['jobID'], $e->getDelay(), $worker->getData(), $e->getName(), $e->getNextRun());
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -349,6 +349,7 @@ try {
                                     'name' => $job['name'],
                                     'id' => $job['jobID'],
                                     'extraParams' => $extraParams,
+                                    'message' => $e->getMessage(),
                                 ]);
                                 try {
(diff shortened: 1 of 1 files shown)


iwiznia (approved):
Nice

bin/BedrockWorkerManager.php
```diff
@@ -349,6 +349,7 @@
                                     'name' => $job['name'],
                                     'id' => $job['jobID'],
                                     'extraParams' => $extraParams,
+                                    'message' => $e->getMessage(),
```
iwiznia, line 352: Oh wait, I think if you pass the exception itself, we will get a better log due to [this](https://github.com/Expensify/Web-Expensify/blob/c729db021a461171810812bf0ca1612ac64d348c/lib/Log.php#L515-L518).

iwiznia (changes requested):

bin/BedrockWorkerManager.php
jasperhuangg: +1

Beamanator: Good catch! I like this better 👍

Beamanator:
Update: Now logging entire `RetryableException` with `'exception' => $e,`

iwiznia (approved):

jasperhuangg (approved):

stitesExpensify (approved):

PR #165, opened 2021-09-16
Sleep less when waiting to queue more jobs so that we gather better data

flodnv:
Explanation here: https://expensify.slack.com/archives/C094TGUTZ/p1631787507041800?thread_ts=1631783921.038500&cid=C094TGUTZ

Tested live. 

Related with https://github.com/Expensify/Expensify/issues/177467

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -173,10 +173,10 @@ try {
                 $stats->counter('bedrockWorkerManager.targetJobsToQueue', $target);
                 break;
             } else {
-                $logger->info('[AIMD] Not enough jobs to queue, waiting 1s and trying again.', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
+                $logger->info('[AIMD] Not enough jobs to queue, waiting 0.5s and trying again.', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
                 $localDB->write('DELETE FROM localJobs WHERE started < '.(microtime(true) - 60 * 60).' AND ended IS NULL;');
                 $isFirstTry = false;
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -173,10 +173,10 @@
                 $stats->counter('bedrockWorkerManager.targetJobsToQueue', $target);
                 break;
             } else {
-                $logger->info('[AIMD] Not enough jobs to queue, waiting 1s and trying again.', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
+                $logger->info('[AIMD] Not enough jobs to queue, waiting 0.5s and trying again.', ['jobsToQueue' => $jobsToQueue, 'target' => $target, 'load' => $load, 'MAX_LOAD' => $maxLoad]);
                 $localDB->write('DELETE FROM localJobs WHERE started < '.(microtime(true) - 60 * 60).' AND ended IS NULL;');
```
flodnv, line 177: We discussed this line with @iwiznia and I stumble upon it again... I think it makes no sense at all. Why would we be deleting jobs that started over an hour ago? They should already be deleted in https://github.com/Expensify/Bedrock-PHP/blob/7e90f8542b8eab2a7f11adc3dd57abca5f387249/bin/BedrockWorkerManager.php#L524. Is it to delete jobs that never end (because of a bug) ??

tylerkaraszewski: https://github.com/Expensify/Bedrock-PHP/pull/100/files#r220564321

Well I know why its `-` instead of `+` now but not why it's there at all. :)

tylerkaraszewski (dismissed):

iwiznia (approved):

flodnv:
Version created and PR to update as well^

iwiznia (approved):

tylerkaraszewski (approved):

PR #168, opened 2021-09-24
Fix warnings caused by previous change

tylerkaraszewski:
This is a fix for the warnings seen here: https://expensify.slack.com/archives/C07J32337/p1632414804283600

As well as some other potential warnings.

Tested locally and verified it doesn't warn.


luacmartins (approved):
LGTM and tests well!

PR #175, opened 2021-12-28
Update to support php 8.1

iwiznia:
Issue https://github.com/Expensify/Expensify/issues/190022


flodnv (commented):
👍 pending comments on your other PRs

flodnv (approved):

.php-cs-fixer.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+// Exclude file inside external libs
```
cead22, line 3:
```suggestion
// Exclude files inside external libs directories
```

.php-cs-fixer.php
```diff
@@ -0,0 +1,33 @@
...
+$finder = PhpCsFixer\Finder::create()
+    ->exclude('vendor')
+    ->exclude('externalLib')
+    ->in(__DIR__)
+    ;
```
cead22, line 8:
```suggestion
    ->in(__DIR__);
```

.php-cs-fixer.php
cead22: Feel free to fix this one as well https://github.com/Expensify/PHP-Libs/pull/476/files#diff-ef867390e21235c28e24595707bf636e1cc9cad29cf8ee4ac405c564a9facdf6R3

.php-cs-fixer.php
cead22: Interesting that the Config is instantiated with new now, but the Finder still uses the factory pattern

cead22 (approved):

PR #182, opened 2022-05-10
Do not wait for BWM children

iwiznia:
# Issue

$ https://github.com/Expensify/Expensify/issues/209619

This does 2 things:
1. Does not wait for children to finish before exiting from the main bwm process. This allows us to re-start a new BWM instance immediately
2. It changes the minimum jobs to dequeue from 6 to 3 (before it had to be `> minSafeJobs / 2` and minSafeJobs is set to 10)

# Tests

- Start BWM
- See it starts some children
- Send the stop signal
- Check BWM main process finishes
- Check the children keep running and finish

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -502,11 +502,7 @@ try {
     echo "Error: $message\r\n";
 }
 
-// We wait for all children to finish before dying.
-$logger->info('Stopping BedrockWorkerManager, waiting for children');
-$status = null;
-pcntl_wait($status);
-$logger->info('Stopped BedrockWorkerManager');
+$logger->info('Stopped BedrockWorkerManager, will not wait for children');
 
 /**
  * Determines whether or not we call GetJob and try to start a new job
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -169,7 +169,9 @@ try {
             if ($load > $maxLoad) {
                 $logger->info('[AIMD] Not safe to start a new job, load is too high, waiting 1s and trying again.', ['load' => $load, 'MAX_LOAD' => $maxLoad]);
                 sleep(1);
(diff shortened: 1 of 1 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -169,7 +169,9 @@
             if ($load > $maxLoad) {
                 $logger->info('[AIMD] Not safe to start a new job, load is too high, waiting 1s and trying again.', ['load' => $load, 'MAX_LOAD' => $maxLoad]);
                 sleep(1);
-            } elseif ($jobsToQueue > $minSafeJobs / 2) {
+            } elseif ($jobsToQueue >= 3) {
```
righdforsa, line 172: NAB: Are we seeing a lot of time where we are waiting for a sixth job? do we really even need to limit it? what’s wrong with if there’s only 1 job to get, getting it? seems like a superfluous throttle at 3 (or for that matter, I’d guess, at 6)

righdforsa (approved):

bin/BedrockWorkerManager.php
iwiznia: I totally agree, we found the origin of the code and it was added because due to aimd, tons of times we were just asking for 1 job then immediately after another job and so on, which has a lot more impact on performance than getting say 5 jobs at a time. So this code was added to help in that. Now I could've deleted it, but my fear was that we were going to go back to getting one job at a time and to avoid that I reduced the number from 6 to 3. 
I would be willing to test removing this completely, but for now this seemed like the safer choice.

iwiznia: Oh just realized I did not answer your first question: yes, we are waiting for the 6th job, that's how I found this issue. You can look [here](https://expensify.slack.com/archives/C03E4EFNGHH/p1651831807894339) (I had more instances, but did not leave links to them, in any case, you should be able to find more by looking at the graphs above that thread)

PR #187, opened 2022-06-09
Re-seed random numbers with each new child PID

johnmlee101:
See this slack thread: https://expensify.slack.com/archives/CQG20GM7X/p1654787619165049

Basically each thread needs a new seed so that they don't produce identical random numbers.


iwiznia (dismissed):

johnmlee101:
Good catch, I always forget Bedrock-PHP doesn't have the auto-release creation

johnmlee101:
k passes and the version exists

iwiznia (approved):

PR #191, opened 2022-11-16
Log when a job takes longer than 1 minute so it's easier to find them

flodnv:
Related with https://github.com/Expensify/Expensify/issues/229197

I did not find a way to find jobs that take forever to run, so I'm adding this log. Tested by adding a sleep in a job and running BWM, and checking logs:
```
2022-11-16T05:36:05.613501-06:00 expensidev2004 php-fpm: ubno3h vendor/bin/BedrockWorkerManager.php we@dont.know !script! ?api? [hmmm] Job took longer than 1 minute ~~ duration: '61.05256485939' name: 'www-prod/ScrapeCards'
```

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -26,7 +26,7 @@ use Expensify\Bedrock\LocalDB;
  */
 
 // Verify it's being started correctly
-if (php_sapi_name() !== "cli") {
+if (php_sapi_name() !== 'cli') {
     // Throw an exception rather than just output because we assume this is
     // being executed on a webserver, so no STDOUT.  Hopefully they've
     // configured a general uncaught exception handler, and this will trigger
@@ -86,7 +86,7 @@ $lastBackoff = 0;
 
 // This is the name of a particular job if it made up over 50% of the jobs previously returned. Its purpose is to
 // detect when we switch job types so that we can react appropriately.
-$lastJobProfile = "none";
+$lastJobProfile = 'none';
 
 $bedrock = Client::getInstance();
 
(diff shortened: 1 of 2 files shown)


bin/BedrockWorkerManager.php
```diff
@@ -469,11 +470,15 @@
...
+                                        $logger->notice('Job took longer than 1 minute', ['name' => $job['name'], 'duration' => $jobDuration]);
+                                    }
                                     $logger->info('Updating local db');
-                                    $stats->benchmark('bedrockWorkerManager.db.write.update', function () use ($localDB, $localJobID) {
-                                        $localDB->write("UPDATE localJobs SET ended=".microtime(true)." WHERE localJobID=$localJobID;");
+                                    $stats->benchmark('bedrockWorkerManager.db.write.update', function () use ($localDB, $localJobID, $time) {
+                                        $localDB->write("UPDATE localJobs SET ended=$time WHERE localJobID=$localJobID;");
                                     });
-                                    $logger->info('Updating local db', ['total' => microtime(true) - $time]);
+                                    $logger->info('Updated local db', ['duration' => microtime(true) - $time]);
```
flodnv, line 481: I was wondering if we could remove the log on this line and line 477 above, they are pretty useless... I left them in doubt 🤷

iwiznia:
I don't think we need this, you can do this all in graphana.

iwiznia:
You can add min and max values to graphs, which would discard anything below/above the value you enter

bin/BedrockWorkerManager.php
```diff
@@ -469,11 +471,15 @@
                             } finally {
                                 if ($enableLoadHandler) {
                                     $time = microtime(true);
+                                    $jobDuration = $time - $jobStartTime;
+                                    if ($jobDuration > 60) {
+                                        $logger->notice('Job took longer than 1 minute', ['name' => $job['name'], 'duration' => $jobDuration]);
```
iwiznia, line 476: Add jobID too?

bin/BedrockWorkerManager.php
iwiznia: I never used them. IIRC Matt B added them at some point to ensure the local DB was not the bottleneck.

iwiznia (dismissed):

flodnv:
Updated

PR #195, opened 2023-08-04
Adding version mismatch handling

danieldoglas:
Partially to solve https://github.com/Expensify/Expensify/issues/170599

We'll start treating version mismatch on servers during deploys. This will blacklist any server with a version mismatch for 5 seconds and let the request be retried in other servers.


danieldoglas:
Closing this in favor of doing it all in bedrock

PR #200, opened 2023-11-21
Decrease default timeout to 120 seconds

flodnv:
Part of https://github.com/Expensify/Expensify/issues/336699, goes with https://github.com/Expensify/Bedrock/pull/1619

Is there anything else I need to modify? In which order should we deploy? I think this should get deployed _after_? 

cc @tylerkaraszewski


src/Client.php
```diff
@@ -240,8 +240,8 @@ public static function configure(array $config)
             'mainHostConfigs' => ['localhost' => ['blacklistedUntil' => 0, 'port' => 8888]],
             'failoverHostConfigs' => ['localhost' => ['blacklistedUntil' => 0, 'port' => 8888]],
             'connectionTimeout' => 1,
-            'readTimeout' => 300,
-            'bedrockTimeout' => 290,
+            'readTimeout' => 120,
+            'bedrockTimeout' => 110,
```
iwiznia, line 244: Ah, so this matches https://github.com/Expensify/Bedrock/pull/1619/files#diff-b8da0476eda8614e5fbb3ab12f6731e1770291753ad84e26605ea64822e0109aR42 and kind of makes that change irrelevant, right?
(still fine to merge)

iwiznia (approved):
Same comment as in the auth PR

src/Client.php
flodnv: I think these values control the PHP timeouts, not the bedrock timeouts?

iwiznia: I think readTimeout does that but bedrockTimeout is the one we pass to bedrock, no?

flodnv: readTimeout is the socket timeout: https://github.com/Expensify/Bedrock-PHP/blob/9f5c7f19c3989acc0a4290907f7b649ca3bc1117/src/Client.php#L533

bedrockTimeout does look like the one we pass: https://github.com/Expensify/Bedrock-PHP/blob/9f5c7f19c3989acc0a4290907f7b649ca3bc1117/src/Client.php#L344-L360

I agree with `still fine to merge`, I am pretty sure we _need_ the change here, and we could have left it alone in Bedrock but better to align both to reduce confusion 😄

PR #203, opened 2024-01-04
Throw BedrockError if we get a 500 response from bedrock

roryabraham:


flodnv:
Closing in favor of https://github.com/Expensify/Bedrock-PHP/pull/206 which is almost the same

PR #204, opened 2024-01-10
Use JSON_PRESERVE_ZERO_FRACTION

aldo-expensify:
PHP when serializing to JSON will strip the `.0` of floats making them look like integers. Then, when we parse these JSONs in Auth, they are parsed as INT instead of FLOAT.

Found this issue investigating: https://github.com/Expensify/App/issues/32824
Asked in slack about it here: https://expensify.slack.com/archives/C03TQ48KC/p1704906489753899

# Test

Check that you can't reproduce the bug in https://github.com/Expensify/App/issues/32824 and verify the distance request is created correctly

cc @johnmlee101 @iwiznia

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -358,7 +358,7 @@ class Client implements LoggerAwareInterface
         $rawRequest = "$method\r\n";
         foreach ($headers as $name => $value) {
             if (is_array($value) || is_object($value)) {
-                $rawRequest .= "$name: ".addcslashes(json_encode($value), '\\')."\r\n";
+                $rawRequest .= "$name: ".addcslashes(json_encode($value, JSON_PRESERVE_ZERO_FRACTION), '\\')."\r\n";
             } elseif (is_bool($value)) {
                 $rawRequest .= "$name: ".($value ? 'true' : 'false')."\r\n";
             } elseif ($value === null || $value === '') {
--- a/src/Client.php
+++ b/src/Client.php
@@ -358,6 +358,9 @@ class Client implements LoggerAwareInterface
         $rawRequest = "$method\r\n";
         foreach ($headers as $name => $value) {
             if (is_array($value) || is_object($value)) {
+                // Passing flag JSON_PRESERVE_ZERO_FRACTION because otherwise PHP will serialize floating numbers like 2.0 to 2 instead of 2.0.
+                // This can cause bugs in Auth since we can't know what type a float number will be and accessing an int with the accessor for
+                // floats (or vice versa) would throw.
                 $rawRequest .= "$name: ".addcslashes(json_encode($value, JSON_PRESERVE_ZERO_FRACTION), '\\')."\r\n";
             } elseif (is_bool($value)) {
                 $rawRequest .= "$name: ".($value ? 'true' : 'false')."\r\n";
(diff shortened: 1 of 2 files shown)


aldo-expensify:
Making draft while testing...

johnmlee101:
Yeah just to be super super duper sure about everything, can you make sure to demo this change on Web-E to make sure it doesn't cause issues 😓

aldo-expensify:
I tested with this: https://github.com/Expensify/App/issues/32824 and it fixes the issue.

> can you make sure to demo this change on Web-E to make sure it doesn't cause issues 😓

What do you mean with "demo this change on Web-E"? like make a video?

johnmlee101:
See that integration tests pass if you set `Bedrock-PHP` to point to this branch in composer

aldo-expensify:
> See that integration tests pass if you set Bedrock-PHP to point to this branch in composer

good idea 👍

aldo-expensify:
Made this branch/PR that uses this version of Bedrock-PHP: https://github.com/Expensify/Web-Expensify/pull/40462
It has a new test creating a distance request that fails if Bedrock-PHP is not updated.

johnmlee101 (approved):
Sweeeeeet. I'm dubiously approving this (Maybe get one more approver?) since I don't know if there's any easy way for us to know which exact flows will be converted this way.

aldo-expensify:
Bumped the version since I think we still have to do it manually

aldo-expensify:
Asked for more eyes in [slack](https://expensify.slack.com/archives/C03TQ48KC/p1704922728722619?thread_ts=1704906489.753899&cid=C03TQ48KC)

src/Client.php
```diff
@@ -358,7 +358,7 @@ public function call($method, $headers = [], $body = '')
         $rawRequest = "$method\r\n";
         foreach ($headers as $name => $value) {
             if (is_array($value) || is_object($value)) {
-                $rawRequest .= "$name: ".addcslashes(json_encode($value), '\\')."\r\n";
+                $rawRequest .= "$name: ".addcslashes(json_encode($value, JSON_PRESERVE_ZERO_FRACTION), '\\')."\r\n";
```
iwiznia, line 361: Can you add a comment explaining this please?

iwiznia (approved):
Looks good, no idea if it will break something, given everything that goes to bedrock passes through this....

src/Client.php
aldo-expensify: How does this sound?

```php
// Passing flag JSON_PRESERVE_ZERO_FRACTION because otherwise PHP will serialize floating numbers like 2.0 to 2 instead of 2.0. This can cause
// bugs in Auth because a 2 is parsed as an INT instead of as a FLOAT.
```

iwiznia: Sounds good, I would change the 2nd sentence to:
> This can cause bugs in Auth since we can't know what type a float number will be and accessing an int with the accessor for floats (or vice versa) would throw

aldo-expensify: Pushed comment

iwiznia (approved):

aldo-expensify:
Going to merge since @johnmlee101 already approved before

aldo-expensify:
Making https://github.com/Expensify/Web-Expensify/pull/40462 ready for review.

PR #215, opened 2024-08-21
Default to null for `port` in $cachedHostConfigs

marcochavezf:
Avoid creating a warning notification if the port is not defined. $cachedHostsAndPorts is still reset if the values don't match cc @flodnv 

$ https://github.com/Expensify/Expensify/issues/420450


flodnv (changes requested):
Why is the port null to begin with? That seems like a bug...

flodnv:
cc @iwiznia

iwiznia (changes requested):
Why would port not be defined? It always should

PR #220, opened 2025-01-10
Update PHP version to 8.3.15 for QSP

grgia:
Part of https://github.com/Expensify/Expensify/issues/459040

I'm not sure about this one because this is the only PR that requires a major bump. @johnmlee101


johnmlee101 (approved):
Oh I must have missed this. However shouldn't affect anything as long as tests pass

cead22 (approved):

PR #229, opened 2025-07-31
Gracefully handle EINPROGRESS

flodnv:
image

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -552,7 +552,29 @@ class Client implements LoggerAwareInterface
             @socket_connect($this->socket, $host, $port);
             $socketErrorCode = socket_last_error($this->socket);
             if ($socketErrorCode === 115) {
-                $this->logger->info('Bedrock\Client - socket_connect returned error 115, continuing.');
+                $this->logger->info('Bedrock\Client - socket_connect returned EINPROGRESS, waiting for connection to complete');
+
+                // Use select to wait for the socket to become writable (indicates connection completion)
+                $write = [$this->socket];
+                $read = null;
+                $except = null;
+
+                $selectResult = socket_select($read, $write, $except, $this->connectionTimeout, $this->connectionTimeoutMicroseconds);
+                if ($selectResult === false) {
+                    $socketError = socket_strerror(socket_last_error());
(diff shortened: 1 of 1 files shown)


src/Client.php
```diff
@@ -552,7 +552,29 @@ private function sendRawRequest(string $host, int $port, string $rawRequest)
...
+                $selectResult = socket_select($read, $write, $except, $this->connectionTimeout, $this->connectionTimeoutMicroseconds);
+                if ($selectResult === false) {
+                    $socketError = socket_strerror(socket_last_error());
+                    throw new ConnectionFailure("Failed to select on socket for host $host:$port. Error: $socketError");
+                } elseif ($selectResult === 0) {
+                    throw new ConnectionFailure("Connection timeout while waiting for EINPROGRESS completion for host $host:$port");
+                }
+
+                // Check if connection completed successfully
+                $connectionError = socket_get_option($this->socket, SOL_SOCKET, SO_ERROR);
```
iwiznia, line 571: Not sure I get this, wouldn't `socket_select` already return an error and that should be handled above?

src/Client.php
```diff
@@ -552,7 +552,29 @@ private function sendRawRequest(string $host, int $port, string $rawRequest)
...
             if ($socketErrorCode === 115) {
-                $this->logger->info('Bedrock\Client - socket_connect returned error 115, continuing.');
+                $this->logger->info('Bedrock\Client - socket_connect returned EINPROGRESS, waiting for connection to complete');
+
+                // Use select to wait for the socket to become writable (indicates connection completion)
+                $write = [$this->socket];
+                $read = null;
+                $except = null;
+
+                $selectResult = socket_select($read, $write, $except, $this->connectionTimeout, $this->connectionTimeoutMicroseconds);
```
iwiznia, line 562: Not that it matters much, but this would basically double the timeout (since we would wait for the timeout in `socket_connect` and then again here.

src/Client.php
flodnv: image

image

image

image

This all sounds right to me 🤔

iwiznia: ok not sure really,  but we can try it

iwiznia (approved):

flodnv:
Version created

iwiznia (approved):

PR #233, opened 2025-09-04
Add logging to debug socket error 11

justinpersaud:
Add logging to see if we can figure out what is going on with these socket errors. More context in https://expensify.slack.com/archives/C094TGUTZ/p1756938582216549

Relates to https://github.com/Expensify/Expensify/issues/544983

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -441,8 +441,8 @@ class Client implements LoggerAwareInterface
                 // Debug EAGAIN errors - check socket state when error occurs
                 $lastSocketError = $this->socket ? socket_last_error($this->socket) : null;
 
-				// Log more data to diagnose problems with EAGAIN/EWOULDBLOCK
-                if ($lastSocketError === 11) { 
+                // Log more data to diagnose problems with EAGAIN/EWOULDBLOCK
+                if ($lastSocketError === 11) {
                     $write = [$this->socket];
                     $read = [];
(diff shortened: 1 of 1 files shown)


src/Client.php
```diff
@@ -438,6 +438,24 @@ public function call($method, $headers = [], $body = '')
...
+                    $write = [$this->socket];
+                    $read = [];
+                    $except = [];
+                    $selectResult = @socket_select($read, $write, $except, 0, 0);
+                    $peerConnected = @socket_getpeername($this->socket, $peerHost, $peerPort);
+
+                    $this->logger->info('EAGAIN Debugging information', [
+                        'host' => $hostName,
+                        'socket_ready_for_writing' => ($selectResult === 1 && !empty($write)) ? 'YES' : 'NO',
+                        'peer_connected' => $peerConnected ? 'YES' : 'NO',
```
iwiznia, line 454: `$peerConnected` can be `0` and `false`, is it ok to log the same in both cases?

iwiznia (changes requested):
Remember that you need to create a tag with the new version. You can use https://github.com/Expensify/Bedrock-PHP/blob/main/package.sh for that

src/Client.php
justinpersaud: Will just check the value of peerConnected instead

PR #239, opened 2025-10-03
Prevent undefined array key access when processing response

francoisl:
### Explanation of Change

When response data doesn't contain any headers (e.g. internal error from remote server) and `extractResponseHeaders()` thus returns an empty array, prevent `$responseHeaders['Content-Length']` from causing a warning for accessing a key that doesn't exist.

### Related Issues
Part of https://github.com/Expensify/Expensify/issues/553941

## Deployment

- [x] I followed the steps in the [README](../README.md#publishing-your-changes) to ensure this PR is deployed properly

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/bedrock-php",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "2.2.2",
+    "version": "2.2.4",
     "authors": [
         {
             "name": "Expensify",


iwiznia (changes requested):
I don't quite get it. We've had millions of requests failing since this code was added, why is this just appearing now? Is this particular command erroring in a different way than all the other "normal" errors?
Also I thought content-length was always set, even for errors. Like in your case `500 Corpay request returned invalid JSON (httpResponseCode: 500,..` that's content and it must have a length, no?

francoisl:
Yes this was it. I saw your message too late but came to the same conclusion 😭 I'll tag you in the Auth PR.

PR #255, opened 2026-06-08
Retry bedrock timeouts better

iwiznia:
# Explanation of Change

This is to fix flakey tests like https://github.com/Expensify/Expensify/issues/610598 and other unrelated tests that fail randomly.
I had claude investigate the failure and what it found makes total sense to me: we retry timeout errors (regardless of idempotency) since when a command times out we did not commit anything, but there are other flows that throw a timeout with different text (`555 Timeout prePeeking command`, `555 Timeout peeking command`, `555 Timeout processing command`, `555 Timeout postProcessing command`) and we are not catching those, which we should. Well, all of them except the postProcessing one, since when we throw that we already committed.


flodnv (approved):
This change makes sense but I don't think this is the main problem of the linked issue, as it seems to happen all the time 😕

iwiznia:
Only one way to know 😄

src/Client.php
```diff
@@ -735,10 +735,16 @@ private function receiveResponse()
...
+        // A command timeout (not a query timeout) is reported with a '555 Timeout' code line, optionally suffixed with
+        // the phase it timed out in ('555 Timeout peeking command', '555 Timeout processing command', etc.). We treat
+        // these as a ConnectionFailure so they get retried regardless of whether the command is idempotent, because the
+        // command timed out before committing and its transaction was rolled back, so no work was applied.
+        // The one exception is '555 Timeout postProcessing command': postProcess runs only after the command has
+        // already committed (200 OK), so retrying it would re-run an already-applied command and could double-apply a
+        // non-idempotent one. We let that case fall through and surface as a regular error.
+        // For any other error code/message we do not throw here, so it gets handled like any regular response.
+        if (str_starts_with($codeLine, '555 Timeout') && $codeLine !== '555 Timeout postProcessing command') {
+            throw new ConnectionFailure("Internal Bedrock command timeout ($codeLine)");
```
aldo-expensify, line 747: I noticed that this affected the timeout error handling in the query timing dashboard. If you time a query that timeout, you throw this new exception. Then, if the requests keeps failing, the exception eventually is converted into a generic Socket exception that doesn't have the original error in here:

https://github.com/Expensify/Web-Expensify/blob/4f36fb30393e7938c43311257eb0469838c54afe/lib/Auth.php#L606-L609

Image

I could fix it for the query timing dashboard adding some code that assumes that a Socket means that the query timeout.. but my concern is if this changed the exception that was bubbling up to the code sending the Auth command, maybe we broken other error handling besides the query timing dashboard with this change.


To reproduce:

1. Hardcode a short timeout in the query timing dashboard (Replace `$selectedTimeoutMS` by `10` [here](https://github.com/Expensify/Web-Expensify/blob/4f36fb30393e7938c43311257eb0469838c54afe/_devportal/tools/timeQuery/index.php#L32))
1. Go to https://www.expensify.com.dev/_devportal/tools/timeQuery/index.php
1. Time a query like `SELECT * FROM reports;`

Debug the exception caught in the `catch (Throwable $e)`

This may be the cause for the CustomQuery being retried when timing queries: https://expensify.slack.com/archives/C05CBC62HGW/p1783010776415689

~Also, I tested with https://github.com/Expensify/Web-Expensify/pull/54229 merged to main, and I think we are still retrying after failing... or that is what I think I'm seeing in the logs.~

aldo-expensify: ah!, I haven't reinstall Bedrock-PHP with composer, so maybe the retry has been fixed... testing again

aldo-expensify: Yeah, I was confused, I saw two CustomQuery commands, but one if for the query plan and the second is the query timing timing out

aldo-expensify: but still, the Exception handling is something we should check

aldo-expensify: Friendly bump @iwiznia

iwiznia: Sorry, had missed this comment.
hmmm so what was the previous error thrown? An ExpException thrown [here](https://github.com/Expensify/Web-Expensify/blob/a7d6c648ed8860b2cb5f125e354f38b83d932497/lib/Auth.php#L186)?

aldo-expensify: Yep! reverted your changes here to verify:

Image

iwiznia: Ah ok, let's catch this exception instead in the time tool? I see we only use the 555 check to change the message a bit but we could just always include the error and the time?

aldo-expensify: but what about other code that may have handled a timeout expecting ExpException and now it doesn't catch it? or we assume such code doesn't exist except for the query timing dashboard?

aldo-expensify: If we just want to fix this at the query dashboard level, here is a PR: https://github.com/Expensify/Web-Expensify/pull/54665

iwiznia: I don't think it exists really, but I guess we will know eventually if it does

PR #256, opened 2026-06-30
Remove retry-everyone-when-down behavior in bedrock client

danieldoglas:
More context https://expensify.slack.com/archives/C05CBC62HGW/p1782740514654479

# Explanation of Change

This PR removes the retry all hosts logic we previously had implemented. This was initially implemented to account for the situations where we were deploying and half of the nodes were unavailable, and the immediately we had the other half unavailable. That would make our logic basically have no hosts for a few seconds.

That doesn't happen anymore because unavailable hosts will forward the request to an available host.

### Related Issues

https://github.com/Expensify/Expensify/issues/649214

## Deployment

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/bedrock-php",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "2.2.10",
+    "version": "2.3.0",
     "authors": [
         {
             "name": "Expensify",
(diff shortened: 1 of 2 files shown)


iwiznia (approved):

iwiznia:
I think you need to re-create the tag because it's pointing to the commit before the merge from main: https://github.com/Expensify/Bedrock-PHP/tree/2.3.0

flodnv (approved):

danieldoglas:
Updated the tag, merging!

PR #257, opened 2026-07-03
Do not blacklist server if maxBlackListTimeout is 0

iwiznia:
# Explanation of Change

Related to https://github.com/Expensify/Web-Expensify/pull/54229 solves this:
> Does not blacklist the host.... well, there's actually an off by one error here, there's 50% chance we blacklist it for 1s [here](https://github.com/Expensify/Bedrock-PHP/blob/main/src/Client.php?rgh-link-date=2026-07-03T11%3A06%3A35Z#L830), LMK if you think I should send a PR to fix that (by doing max(rand(1, $this->maxBlackListTimeout), $this->maxBlackListTimeout))

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -830,8 +830,7 @@ class Client implements LoggerAwareInterface
         if (!$this->maxBlackListTimeout) {
             return;
         }
-        $blacklistLength = rand(1, $this->maxBlackListTimeout);
-        $blacklistedUntil = time() + $blacklistLength;
+        $blacklistedUntil = time() + rand(1, $this->maxBlackListTimeout);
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -827,7 +827,11 @@ private static function toUTF8($str)
      */
     private function markHostAsFailed(string $host)
     {
-        $blacklistedUntil = time() + rand(1, $this->maxBlackListTimeout);
+        $blacklistLength = min(rand(1, $this->maxBlackListTimeout), $this->maxBlackListTimeout);
```
flodnv: Sort of strange way to do this... could we just return early `if (!$this->maxBlackListTimeout)`?

iwiznia: Sure

flodnv (approved):

PR #258, opened 2026-07-09
Add a circuit breaker to the Bedrock client

justinpersaud:
# Explanation of Change

Adds a circuit breaker to the client.

The way it works is we wrap the existing call method with the circuit breaker logic. If 10 requests fail, we fail requests immediately for the next 10 seconds and then check if they're working again. If they work, close the breaker and resume normal operation. If it fails, keep it open and wait another 10 seconds.

### Related Issues

Relates to https://github.com/Expensify/Expensify/issues/657702

It's an alternate way of solving the issue.

## Deployment

## Tests

Claude wrote a test script for me. It 
* sends requests to auth in my dev enviroment and confirms they work
* sends requests to an invalid port where no bedrock is listening and confirms it trips a circuit breaker, but uses the same cluster name prefix to fake trip it
* sends requests to auth again and confirm the circuit breaker closed and they work

```
<?php
/**
 * Manual test for the Bedrock circuit breaker (Bedrock-PHP #258).
 *
 * Proves three behaviors in one run, and prints the underlying APCu breaker state after every call:
 *   1. HEALTHY   — normal traffic is unaffected (breaker stays closed; fails=none).
 *   2. DOWN      — failures make the fails counter climb to the threshold, the breaker opens, then
 *                  subsequent calls fail fast.
 *   3. RECOVERY  — after the cooldown a healthy probe succeeds and both APCu keys are cleared.
 *
 * It does NOT stop or mock Bedrock: healthy phases hit the real local Bedrock (8888); the "down"
 * phase just points the client at a dead port (connection refused).
 *
 * Run inside the Expensidev VM (APCu on for CLI; auto_prepend off so the app bootstrap doesn't load
 * the vendor Client before we load the branch one):
 *   php -d apc.enable_cli=1 -d auto_prepend_file= -d auto_append_file= /tmp/circuit-breaker-test.php
 *
 * /vagrant/Bedrock-PHP must be on the PR branch (jpersaud-bedrock-circuit-breaker).
 */

require '/vagrant/Web-Expensify/vendor/autoload.php'; // deps: PSR log, NullStats, exceptions
require '/vagrant/Bedrock-PHP/src/Client.php';        // load the breaker version of the client

use Expensify\Bedrock\Client;

const CLUSTER = 'cbmanualtest'; const THRESHOLD = 3; const COOLDOWN = 5;
const HEALTHY = 8888;   // real local Bedrock
const DEAD_PORT = 9997; // nothing listening -> connection refused
const CMD = 'Status';

if (!defined('Expensify\\Bedrock\\Client::CIRCUIT_BREAKER_CACHE_PREFIX')) { fwrite(STDERR, "ERROR: no breaker — is /vagrant/Bedrock-PHP on the PR branch?\n"); exit(1); }
if (!(function_exists('apcu_enabled') && apcu_enabled())) { fwrite(STDERR, "ERROR: APCu off for CLI — run with -d apc.enable_cli=1\n"); exit(1); }

$failsKey = Client::CIRCUIT_BREAKER_CACHE_PREFIX . CLUSTER . '-fails';
$openKey  = Client::CIRCUIT_BREAKER_CACHE_PREFIX . CLUSTER . '-open';
apcu_delete($failsKey); apcu_delete($openKey); // clean start

// Reads the live APCu breaker state so we can display it as proof.
$state = function () use ($failsKey, $openKey): string {
    $f = apcu_fetch($failsKey);
    $o = apcu_fetch($openKey);
    $fails = ($f === false) ? 'none' : (string) $f;
    if ($o === false) { $open = 'none'; }
    else { $left = $o - time(); $open = $left > 0 ? "OPEN({$left}s left)" : 'cooldown-elapsed'; }
    return sprintf('apcu[fails=%-4s open=%s]', $fails, $open);
};

$make = fn(int $port) => Client::getInstance([
    'clusterName' => CLUSTER, 'logger' => new \Psr\Log\NullLogger(), 'stats' => new \Expensify\Bedrock\Stats\NullStats(),
    'mainHostConfigs' => ['127.0.0.1' => ['blacklistedUntil' => 0, 'port' => $port]], 'failoverHostConfigs' => [],
    'connectionTimeout' => 0, 'connectionTimeoutMicroseconds' => 500000, 'readTimeout' => 3, 'readTimeoutMicroseconds' => 0,
    'bedrockTimeout' => 3, 'maxBlackListTimeout' => 1, 'writeConsistency' => 'ASYNC', 'commandPriority' => null, 'logParam' => null,
    'circuitBreakerThreshold' => THRESHOLD, 'circuitBreakerCooldown' => COOLDOWN,
]);
$call = function (Client $c): array {
    $t = microtime(true);
    try { $c->call(CMD); return ['OK', (microtime(true) - $t) * 1000, '']; }
    catch (\Throwable $e) {
        $ff = strpos($e->getMessage(), 'circuit breaker open') !== false;
        return [$ff ? 'FAIL-FAST' : 'FAIL', (microtime(true) - $t) * 1000, $ff ? 'circuit breaker open' : trim(substr($e->getMessage(), 0, 42))];
    }
};

printf("======================================================================\n");
printf(" Bedrock circuit-breaker test — %s\n cluster=%s  threshold=%d  cooldown=%ds  apcu=ON\n", date('Y-m-d H:i:s'), CLUSTER, THRESHOLD, COOLDOWN);
printf("======================================================================\n\n");
$pass = 0;

printf("PHASE 1 — HEALTHY (real Bedrock 127.0.0.1:%d)\n", HEALTHY);
$h = $make(HEALTHY); $ok = true;
for ($i = 1; $i <= 5; $i++) { [$r, $ms] = $call($h); $ok = $ok && $r === 'OK'; printf("      [%s] call %d: %-9s %6.1f ms   %s\n", date('H:i:s'), $i, $r, $ms, $state()); }
$ok = $ok && apcu_fetch($openKey) === false; $pass += $ok; printf("   => breaker CLOSED, all calls succeeded, no APCu state   [%s]\n\n", $ok ? 'PASS' : 'FAIL');

printf("PHASE 2 — CLUSTER DOWN (dead port 127.0.0.1:%d)\n", DEAD_PORT);
$d = $make(DEAD_PORT); $sawFF = false; $before = 0; $tripped = false;
for ($i = 1; $i <= 6; $i++) {
    [$r, $ms, $note] = $call($d);
    if ($r === 'FAIL-FAST') { $sawFF = true; $tripped = true; } elseif (!$tripped) { $before++; }
    printf("      [%s] call %d: %-9s %6.1f ms  (%-22s) %s%s\n", date('H:i:s'), $i, $r, $ms, $note, $state(), $r === 'FAIL-FAST' ? '  <-- instant' : '');
}
$ok = $sawFF && $before === THRESHOLD && (int) apcu_fetch($failsKey) >= THRESHOLD; $pass += $ok;
printf("   => fails climbed to %d, breaker OPENED, then failed fast   [%s]\n\n", $before, $ok ? 'PASS' : 'FAIL');

printf("PHASE 3 — RECOVERY (wait %ds past cooldown, then real Bedrock)\n", COOLDOWN + 1);
printf("      [%s] sleeping %ds...   %s\n", date('H:i:s'), COOLDOWN + 1, $state());
sleep(COOLDOWN + 1);
$rc = $make(HEALTHY); $ok = true;
for ($i = 1; $i <= 3; $i++) { [$r, $ms] = $call($rc); $ok = $ok && $r === 'OK'; printf("      [%s] call %d: %-9s %6.1f ms   %s%s\n", date('H:i:s'), $i, $r, $ms, $state(), $i === 1 ? '  <-- probe OK, state cleared' : ''); }
$ok = $ok && apcu_fetch($failsKey) === false && apcu_fetch($openKey) === false; $pass += $ok;
printf("   => probe succeeded, both APCu keys cleared, service resumed   [%s]\n\n", $ok ? 'PASS' : 'FAIL');

printf("======================================================================\n RESULT: %d/3 phases PASS — circuit breaker %s\n======================================================================\n", $pass, $pass === 3 ? 'works as expected' : 'FAILED');
exit($pass === 3 ? 0 : 1);
```

```
expensidev-2404->/vagrant/Bedrock-PHP (jpersaud-bedrock-circuit-breaker) > php -d apc.enable_cli=1 -d auto_prepend_file= -d auto_append_file= /tmp/circuit-breaker-test.php
======================================================================
 Bedrock circuit-breaker test — 2026-07-13 17:34:50
 cluster=cbmanualtest  threshold=3  cooldown=5s  apcu=ON
======================================================================

PHASE 1 — HEALTHY (real Bedrock 127.0.0.1:8888)
      [17:34:50] call 1: OK           0.5 ms   apcu[fails=none open=none]
      [17:34:50] call 2: OK           0.3 ms   apcu[fails=none open=none]
      [17:34:50] call 3: OK           0.2 ms   apcu[fails=none open=none]
      [17:34:50] call 4: OK           0.2 ms   apcu[fails=none open=none]
      [17:34:50] call 5: OK           0.2 ms   apcu[fails=none open=none]
   => breaker CLOSED, all calls succeeded, no APCu state   [PASS]

PHASE 2 — CLUSTER DOWN (dead port 127.0.0.1:9997)
      [17:34:50] call 1: FAIL         1.0 ms  (Failed to send request to bedrock host 127) apcu[fails=1    open=none]
      [17:34:50] call 2: FAIL         0.1 ms  (Failed to send request to bedrock host 127) apcu[fails=2    open=none]
      [17:34:50] call 3: FAIL         0.0 ms  (Failed to send request to bedrock host 127) apcu[fails=3    open=OPEN(5s left)]
      [17:34:50] call 4: FAIL-FAST    0.0 ms  (circuit breaker open  ) apcu[fails=3    open=OPEN(5s left)]  <-- instant
      [17:34:50] call 5: FAIL-FAST    0.0 ms  (circuit breaker open  ) apcu[fails=3    open=OPEN(5s left)]  <-- instant
      [17:34:50] call 6: FAIL-FAST    0.0 ms  (circuit breaker open  ) apcu[fails=3    open=OPEN(5s left)]  <-- instant
   => fails climbed to 3, breaker OPENED, then failed fast   [PASS]

PHASE 3 — RECOVERY (wait 6s past cooldown, then real Bedrock)
      [17:34:50] sleeping 6s...   apcu[fails=3    open=OPEN(5s left)]
      [17:34:56] call 1: OK           1.3 ms   apcu[fails=none open=none]  <-- probe OK, state cleared
      [17:34:56] call 2: OK           0.8 ms   apcu[fails=none open=none]
      [17:34:56] call 3: OK           1.1 ms   apcu[fails=none open=none]
   => probe succeeded, both APCu keys cleared, service resumed   [PASS]

======================================================================
 RESULT: 3/3 phases PASS — circuit breaker works as expected
======================================================================
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -296,9 +296,9 @@ class Client implements LoggerAwareInterface
             'stats' => new NullStats(),
             'writeConsistency' => 'ASYNC',
             'maxBlackListTimeout' => 1,
-            'circuitBreakerThreshold' => 0,
-            'circuitBreakerCooldown' => 5,
-            'retryBudgetRatio' => 0,
+            'circuitBreakerThreshold' => 10,
+            'circuitBreakerCooldown' => 10,
+            'retryBudgetRatio' => 0.1,
             'commandPriority' => null,
             'logParam' => null,
         ], self::$defaultConfig, $config);
--- a/src/Client.php
+++ b/src/Client.php
@@ -347,9 +347,6 @@ class Client implements LoggerAwareInterface
      * @param string $body    Request body (optional)
      *
      * @return array JSON response
-     *
-     * @throws BedrockError
-     * @throws ConnectionFailure
      */
     public function call($method, $headers = [], $body = '')
     {
--- a/src/Client.php
+++ b/src/Client.php
@@ -30,16 +30,9 @@ class Client implements LoggerAwareInterface
     public const APCU_CACHE_PREFIX = 'bedrockHostConfigs-';
 
     /**
-     * APCu key prefixes for the (per-cluster, box-wide) circuit breaker and retry-budget state.
+     * APCu key prefix for the (per-cluster, box-wide) circuit breaker state.
      */
     public const CIRCUIT_BREAKER_CACHE_PREFIX = 'bedrockCircuitBreaker-';
-    public const RETRY_BUDGET_CACHE_PREFIX = 'bedrockRetryBudget-';
-
-    /**
-     * Attempts needed in the current minute before the retry-budget ratio is enforced (avoids a cold-start
-     * where a single early retry trips the ratio).
-     */
-    public const RETRY_BUDGET_MIN_SAMPLE = 100;
 
     /**
      * Priorities a command can have.
@@ -153,11 +146,6 @@ class Client implements LoggerAwareInterface
     private $circuitBreakerCooldown;
 
     /**
-     * @var float Max retries as a fraction of total attempts per cluster per minute (box-wide). 0 disables it.
-     */
-    private $retryBudgetRatio;
-
-    /**
      * @var bool Set this to true to add a `mockRequest` header to all outgoing requests.
      */
     private $mockRequests;
@@ -198,7 +186,6 @@ class Client implements LoggerAwareInterface
      *                      string|null          logParam            Extra data to add to the bedrock logs
      *                      int|null             circuitBreakerThreshold Cluster-unreachable failures before the breaker opens (0 disables)
      *                      int|null             circuitBreakerCooldown  Seconds the breaker stays open once tripped
-     *                      float|null           retryBudgetRatio        Max retries as a fraction of attempts/cluster/minute (0 disables)
      *
      * @throws BedrockError
      */
@@ -219,7 +206,6 @@ class Client implements LoggerAwareInterface
         $this->maxBlackListTimeout = $config['maxBlackListTimeout'];
         $this->circuitBreakerThreshold = $config['circuitBreakerThreshold'];
         $this->circuitBreakerCooldown = $config['circuitBreakerCooldown'];
-        $this->retryBudgetRatio = $config['retryBudgetRatio'];
         $this->commandPriority = $config['commandPriority'];
         $this->logParam = $config['logParam'];
 
@@ -298,7 +284,6 @@ class Client implements LoggerAwareInterface
             'maxBlackListTimeout' => 1,
             'circuitBreakerThreshold' => 10,
             'circuitBreakerCooldown' => 10,
-            'retryBudgetRatio' => 0.1,
             'commandPriority' => null,
             'logParam' => null,
         ], self::$defaultConfig, $config);
@@ -479,16 +464,7 @@ class Client implements LoggerAwareInterface
         }
 
         $hostName = null;
-        $attempt = 0;
         while (!$response && count($hostConfigs)) {
-            // Every attempt after the first is a retry; a box-wide retry budget keeps a cluster-wide
-            // outage from turning each request into a burst of cross-host retries (retry amplification).
-            if ($attempt > 0 && !$this->consumeRetryBudget()) {
-                $this->logger->info('Bedrock\Client - Retry budget exhausted, not retrying', ['clusterName' => $this->clusterName]);
-                break;
-            }
-            $this->recordAttempt();
-            $attempt++;
             $exception = null;
             reset($hostConfigs);
             $numRetriesLeft = count($hostConfigs) - 1;
@@ -978,41 +954,6 @@ class Client implements LoggerAwareInterface
         }
         apcu_delete(self::CIRCUIT_BREAKER_CACHE_PREFIX.$this->clusterName.'-fails');
         apcu_delete(self::CIRCUIT_BREAKER_CACHE_PREFIX.$this->clusterName.'-open');
-    }
-
-    /**
-     * Counts one attempt (first try or retry) toward the current minute's box-wide retry-budget window.
-     */
-    private function recordAttempt(): void
-    {
-        if ($this->retryBudgetRatio <= 0 || !$this->apcuAvailable()) {
-            return;
-        }
-        $success = false;
-        apcu_inc(self::RETRY_BUDGET_CACHE_PREFIX.$this->clusterName.'-att-'.floor(time() / 60), 1, $success, 120);
-    }
-
-    /**
-     * Enforces a box-wide retry budget: retries are allowed only while they stay under retryBudgetRatio of
-     * total attempts this minute, capping retry amplification during an outage without throttling normal
-     * failover. Returns true when a retry is permitted.
-     */
-    private function consumeRetryBudget(): bool
-    {
-        if ($this->retryBudgetRatio <= 0 || !$this->apcuAvailable()) {
-            return true;
-        }
-        $minute = floor(time() / 60);
-        $attempts = apcu_fetch(self::RETRY_BUDGET_CACHE_PREFIX.$this->clusterName.'-att-'.$minute) ?: 0;
-        $success = false;
-        $retries = apcu_inc(self::RETRY_BUDGET_CACHE_PREFIX.$this->clusterName.'-ret-'.$minute, 1, $success, 120);
-
-        // Allow retries freely until there's a meaningful sample this minute; low traffic can't storm anyway.
-        if ($attempts < self::RETRY_BUDGET_MIN_SAMPLE) {
-            return true;
-        }
-
-        return ($retries / $attempts) <= $this->retryBudgetRatio;
     }
 
     /**
--- a/src/Client.php
+++ b/src/Client.php
@@ -30,6 +30,18 @@ class Client implements LoggerAwareInterface
     public const APCU_CACHE_PREFIX = 'bedrockHostConfigs-';
 
     /**
+     * APCu key prefixes for the (per-cluster, box-wide) circuit breaker and retry-budget state.
+     */
+    public const CIRCUIT_BREAKER_CACHE_PREFIX = 'bedrockCircuitBreaker-';
+    public const RETRY_BUDGET_CACHE_PREFIX = 'bedrockRetryBudget-';
+
+    /**
+     * Attempts needed in the current minute before the retry-budget ratio is enforced (avoids a cold-start
+     * where a single early retry trips the ratio).
+     */
+    public const RETRY_BUDGET_MIN_SAMPLE = 100;
+
+    /**
      * Priorities a command can have.
      */
     public const PRIORITY_MIN = 0;
@@ -131,6 +143,21 @@ class Client implements LoggerAwareInterface
     private $maxBlackListTimeout;
 
     /**
+     * @var int Consecutive cluster-unreachable failures before the circuit breaker opens. 0 disables it.
+     */
+    private $circuitBreakerThreshold;
+
+    /**
+     * @var int Seconds the circuit breaker stays open (failing fast) once tripped.
+     */
+    private $circuitBreakerCooldown;
+
+    /**
+     * @var float Max retries as a fraction of total attempts per cluster per minute (box-wide). 0 disables it.
+     */
+    private $retryBudgetRatio;
+
+    /**
      * @var bool Set this to true to add a `mockRequest` header to all outgoing requests.
      */
     private $mockRequests;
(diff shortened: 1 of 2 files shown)


danieldoglas (dismissed):

src/Client.php
```diff
@@ -169,6 +184,8 @@ class Client implements LoggerAwareInterface
      *                      int|null             maxBlackListTimeout When a host fails, it will blacklist it and not try to reuse it for up to this amount of seconds.
      *                      int|null             commandPriority     The priority to send the commands with
      *                      string|null          logParam            Extra data to add to the bedrock logs
+     *                      int|null             circuitBreakerThreshold Cluster-unreachable failures before the breaker opens (0 disables)
+     *                      int|null             circuitBreakerCooldown  Seconds the breaker stays open once tripped
```
iwiznia: Why are we allowing `null` here?

src/Client.php
```diff
@@ -846,6 +900,84 @@ private function markHostAsFailed(string $host)
         }
     }
 
+    /**
+     * Whether APCu-backed shared state (the circuit breaker) is usable in this environment.
+     *
+     * @suppress PhanUndeclaredConstant - suppresses ARE_GITHUB_ACTIONS_RUNNING
+     */
+    private function apcuAvailable(): bool
```
iwiznia:
```suggestion
    private function isApcuAvailable(): bool
```

src/Client.php
```diff
@@ -846,6 +900,84 @@ private function markHostAsFailed(string $host)
...
+    /**
+     * Records a cluster-unreachable failure.
+     *
+     * If the breaker is already tripped (the '-open' key exists — i.e. we're in the post-cooldown probe
+     * window), a single failure re-opens it immediately and refreshes that key, so a still-down cluster
+     * stays tripped for the whole outage without having to re-accumulate the threshold. Otherwise the
+     * breaker is closed and we count failures toward the threshold with an atomic apcu_inc (so concurrent
+     * workers can't lose increments).
+     *
+     * TTLs are cooldown + 60s: long enough that the '-open' key always outlives its cooldown window (so
```
iwiznia, line 930: Not sure I get this, why have a cooldown config if we always add 60s to it? Why not let the consumer set the cooldown to 60s more if that's what they want?

iwiznia: Oh, this is for the apcu store call 👍

src/Client.php
justinpersaud: copy and paste from other methods, going to drop it since we only use int

src/Client.php
justinpersaud: Yeah and to clarify, the TTL is a fail safe.

The breaker only clears if it sends a new request that succeeds once the cooldown has expired. In theory, the server could not receive a request for whatever reason and then nothing would clear it this way, thus the TTL would expire and it would clear on it's own. In theory, this will probably never happen, but there is no harm in having it here.

iwiznia (dismissed):
NAB but just realized you could use `isApcuAvailable` in `getPossibleHosts` and `markHostAsFailed` that have similar logic

justinpersaud:
good point, will apply that

iwiznia (dismissed):

danieldoglas (dismissed):

src/Client.php
```diff
@@ -314,6 +335,39 @@ public function getClusterName(): string
...
+     *
+     * @return array JSON response
+     *
+     * @throws BedrockError
+     * @throws ConnectionFailure
+     */
+    public function call($method, $headers = [], $body = '')
+    {
+        if (!$this->circuitBreakerAllowsRequest()) {
+            $this->logger->info('Bedrock\Client - Circuit breaker open, failing fast', ['clusterName' => $this->clusterName]);
```
danieldoglas, line 354: this will log thousands of times when it is open. are we worried about that?

justinpersaud: It will, but I think it's actually less than before this change where we log the "Marking server as failed"

The alternative is to remove this, but then that makes follow up troubleshooting more difficult

danieldoglas: Won't the throw below already be logged too?

justinpersaud: Hm yes good point. So all the callers that catch that exception will probably be a bit noisy, though I'm not sure if it's any more than it is today. Should we wait and see perhaps?

src/Client.php
```diff
@@ -314,6 +335,39 @@ public function getClusterName(): string
...
+     * @param string $method  Request method
+     * @param array  $headers Request headers (optional)
+     * @param string $body    Request body (optional)
+     *
+     * @return array JSON response
+     *
+     * @throws BedrockError
+     * @throws ConnectionFailure
+     */
+    public function call($method, $headers = [], $body = '')
```
flodnv, line 351: Add types? Maybe?

src/Client.php
```diff
@@ -314,6 +335,39 @@ public function getClusterName(): string
...
+     * they never trip it.
+     *
+     * @param string $method  Request method
+     * @param array  $headers Request headers (optional)
+     * @param string $body    Request body (optional)
+     *
+     * @return array JSON response
+     *
+     * @throws BedrockError
+     * @throws ConnectionFailure
```
flodnv: NAB these aren't necessary

src/Client.php
flodnv: I think this is fine, but shouldn't this be an alert?

flodnv (commented):
NABs

src/Client.php
iwiznia: If you do, you need to ensure it's safe for all callers....

src/Client.php
iwiznia: I don't think so? Will create an issue when it's open and there's not much to do to fix it other than recover the cluster which we should already be working on by the time the issue is created, no?

src/Client.php
justinpersaud: that sounds like a can of worms

justinpersaud:
Got rid of the throws, omitting the other ones for now. Need a final approval

iwiznia (approved):

danieldoglas (approved):

src/Client.php
flodnv: My thought was more about seeing this problem in logs... we could whitelist this to not create GH issues, which is something I've wanted to do with other warnings/alerts that I still want to see in logwatcher [like this one](https://github.com/Expensify/Expensify/issues/647310#issuecomment-4923013371)

iwiznia:
Should we be skipping the circuit breaker for things like marqeta related calls? Specially the ones related to authorizations, as the webhook will get retried anyway

Similar comment to https://github.com/Expensify/Auth/pull/22871#pullrequestreview-4715903606

justinpersaud:
Hmm, I don't follow. If the command is queued and has been timing out because the cluster is down or something is broken, does it make sense to continuously try and send it? We're just going to fail it without sending it at all if the breaker is open

iwiznia:
That the breaker is open does not necessarily mean all requests will fail if we attempted them, right? Cluster might have recovered already for example...

justinpersaud:
When the breaker is open, all requests will fail when attempted unless the cooldown has expired and the next request is the probing one, in which case we attempt to see if it works, and if it does, close the breaker and resume as normal.

iwiznia:
Yeah, I get that, but the breaker is preemptively failing requests because we think they will fail. There's a chance the cluster recovered or that a particular request does succeed if we actually sent it.
And my point is that maybe we should allow requests from the gateway to be sent and get attempted. The volume of them is low enough to not matter much in terms of impact on the servers and we don't fallback to the CF worker which I think still has less logic on what to reject than the auth command.

justinpersaud:
IMO this is a bit of a slippery slope. The whole point of this change is to allow the cluster to recover, and once we start making exceptions, we can put ourselves in the same situation this change aims to prevent.

For the marqeta gateway case itself, I think that is better handled by our cloudflare worker for marqeta-gateway which is the whole purpose of that thing.

As it stands right now, what should happen when the breaker is open:

* We receive authorization request from the webhook
* `GetExpensifyCardAvailableSpend` receives a socket error from Auth because the breaker is open
* gateway.php returns a 500 error which sends it to commando mode

iwiznia:
ok sounds good

flodnv:
I agree

src/Client.php
flodnv: Bump on this question. How are we going to know in logwatcher that this is happening? Will there be another alert further up the call stack?

justinpersaud:
Going to merge and get a PR up for Web-E

src/Client.php
flodnv: @justinpersaud ^

iwiznia:
Don't forget web-s

justinpersaud:
https://github.com/Expensify/Web-Expensify/pull/54637
https://github.com/Expensify/Web-Secure/pull/2987

src/Client.php
justinpersaud: lets change it to warn then?

@MelvinBot create a new github issue for this and assign it to me

PR #263, opened 2026-07-27
[WIP] Add per-type AIMD load handling to BedrockWorkerManager

chiragsalian:
# Explanation of Change

Per-type AIMD for the Bedrock Worker Manager, so a single slow or flooding job type can no longer take down shared backends (context: the Uber/Nutanix self-DDOS, [Expensify#661712](https://github.com/Expensify/Expensify/issues/661712)).

- Extracts the existing scalar AIMD out of `BedrockWorkerManager.php` into a pure, unit-testable `AimdController` (behavior-preserving), and adds a PHPUnit harness + CI job (the repo had no tests).
- Adds a flag-gated (`--enablePerTypeAimd`) per-type handler that tracks a target **per job type**, so a slow type backs off on its own latency instead of being masked by fast jobs.
- Wires up the previously-dead `--maxSafeTime` as an **absolute** latency backoff (with ramp-up suppression), plus a low per-type floor and protected floors for critical jobs (`--criticalTypeFloors`) and per-type thresholds (`--maxSafeTimeOverrides`), supplied as JSON so app-specific job names stay out of this generic library.
- Adds `LocalDB::readAll()` for the per-type `GROUP BY` reads.
- 16 unit tests cover the per-type behavior.

**Off by default:** the scalar handler stays the behavior until `--enablePerTypeAimd` is set, so this is safe to land and roll out gradually.

## What this PR solves / does not solve

**Solves — the system now NOTICES the bad job type early.** Old (scalar) AIMD averaged all jobs together: 5,000 fast jobs kept the average looking healthy, so it never realized `SendFetchableOnyxUpdates` was the problem and kept pulling it at full speed (even ramping up). It was blind. Per-type AIMD measures `SendFetchableOnyxUpdates` against its own history; its average shoots past 30s, so we back that type's target down toward the floor and stop it from ramping — while leaving the fast jobs alone. We identify the culprit and stop actively feeding it more budget within ~10–20 seconds instead of never. That's the exact failure from the fire, fixed.

**Does NOT solve — it can't stop jobs already in the queue from slipping through.** After backing `SendFetchableOnyxUpdates` down to "run ~1":

- The healthy fast jobs open a fetch budget of, say, 35 jobs.
- We ask Bedrock for "35 jobs, any type." But the queue holds 10,000 heavy jobs, so Bedrock hands back 35 heavy ones, and we launch all 35.
- We wanted 1; we ran 35.

So this PR makes the throttle a **soft** limit (it shrinks the total we pull), not a **hard** cap (it can't refuse the bad type specifically). The hard cap needs the fork-time admission gate (a follow-up PR): after those 35 come back, requeue 34 of them instead of running them.

| | This PR | Needs admission gate (follow-up) |
|---|---|---|
| Notice a slow type despite fast jobs masking it | ✅ | |
| Stop ramping up the bad type | ✅ | |
| Shrink the total jobs pulled during a flood | ✅ (partial) | |
| Hard cap on how many of the bad type run at once | | ❌ |

## Test Steps
1. I created a bedrock Job called SlowTestJob. It has nothing other than a log.
2. I queued 100 of these jobs. Ran bwm. Tailed the logs for `Congestion Avoidance for type`.
3. Confirmed i saw `Congestion Avoidance for type` for the job and it started ramping up,
```
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '1.5034158229828'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '1.5034158229828'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '1.5407197475433'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '1.5548987388611'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '2.0630359649658'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '2.5644588470459'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '3.0725469589233'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '3.5750317573547'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '3.5750317573547'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '3.6077847480774'
[info] [AIMD] Congestion Avoidance for type. ~~ type: 'SlowTestJob' target: '3.6146068572998'
```
So this shows up the congestion was detecting my job specifically, even even though other job types were there locally. We see from its target that it's ramping up, and it ended up with 3.6, meaning that at the time BWM safely thought it could run 3.6 of these jobs at a time. 

4. I updated SlowTestJob with a sleep of 40s. 
5. I queued 100 of these jobs. Ran bwm. Tailed the logs for `Backing off type target`.
6. Confirm that I saw this in the log as soon as the first set of jobs completed,
```
[info] [AIMD] Backing off type target. ~~ type: 'SlowTestJob' target: '1' lastIntervalAverage: '40.021671431405' reason: 'maxSafeTime'
```
This indicates that BWM believes that only one job of this can be picked up at a time, which is the current default minimum. It also shows us that the average time to complete one of the jobs is 40 seconds. 

### Related Issues
For https://github.com/Expensify/Expensify/issues/661712

## Deployment

### DIFF
--- a/.github/workflows/bedrock-php.yml
+++ b/.github/workflows/bedrock-php.yml
@@ -66,3 +66,30 @@ jobs:
 
     - name: Test for Style
       run: "php ./ci/style.php"
+  PHP_Unit:
+    name: PHPUnit
+    runs-on: blacksmith-2vcpu-ubuntu-2404
+    timeout-minutes: 15
+    steps:
+    - name: checkout
+      # v4.1.0
+      uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608
+      with:
+        # Set fetch-depth to 100 so that we can compare HEAD with a good chunk of git log history
+        fetch-depth: 100
+
+    - name: Install PHP and Libraries
+      uses: shivammathur/setup-php@9e72090525849c5e82e596468b86eb55e9cc5401 # v2.32.0
+      with:
+        tools: composer:v2.7.1
+        php-version: 8.3.31
+        coverage: none
+      env:
+        GITHUB_TOKEN: ${{ secrets.CODE_EXPENSIFY_TOKEN }}
+        fail-fast: true
+
+    - name: Setup Composer Cache
+      uses: ./.github/actions/composite/setup-composer-cache
+
+    - name: Run PHPUnit tests
+      run: "vendor/bin/phpunit"
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -2,6 +2,12 @@
 
 declare(strict_types=1);
 
+use Expensify\Bedrock\Aimd\AimdConfig;
+use Expensify\Bedrock\Aimd\AimdController;
+use Expensify\Bedrock\Aimd\AimdIntervalStats;
+use Expensify\Bedrock\Aimd\PerTypeAimdConfig;
+use Expensify\Bedrock\Aimd\PerTypeAimdController;
+use Expensify\Bedrock\Aimd\PerTypeIntervalStats;
 use Expensify\Bedrock\Client;
 use Expensify\Bedrock\Exceptions\BedrockError;
 use Expensify\Bedrock\Exceptions\Jobs\DoesNotExist;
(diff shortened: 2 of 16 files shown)


chiragsalian:
Closing — pivoting away from BWM-side pacing to a server-side fix that caps GetOnyxUpdatesForAsyncNotification thread fanout directly in Auth (see Expensify/Auth#23365).

PR #265, opened 2026-07-28
[WIP] Add backend-health gate to BWM

chiragsalian:
# Explanation of Change

Adds the **backend-health gate** to the BWM — the durable half of the AIMD fix. Per-type AIMD (#263) is a *local, relative* pacer; this adds an *absolute* signal tied to the resource that actually saturates, so BWM stops pulling work when the backend is in trouble instead of watching the local box's loadavg (which stayed ~26/48 "Safe" throughout the 2026-07-08 fire while Auth CPU was pinned at ~100%).

- `Status::getHealth()` / `parseHealth()` — reads the backend's in-flight `commandCount` and queue depth from the `Status` command. `parseHealth()` is pure and unit-tested.
- Top-of-loop gate — pauses fetching when `commandCount > --maxBackendCommandCount` (**`0` = disabled, the default**). Cached poll (`--healthCheckIntervalSeconds`, default 2s); **fails closed** (treats an unreachable backend as saturated so we don't pile onto a dying node). The local `sys_getloadavg()` check is kept as a secondary guard.
- 4 new tests for `parseHealth`; full suite green (24 tests).

**Off by default** — merging changes nothing until `--maxBackendCommandCount` is set in Salt.

## Stacking / dependency

Based on the **#263 branch** (`aimd-extract-testable-controller`), so this diff shows only the gate. Retarget to `main` once #263 merges. Ships in the same bedrock-php line as per-type AIMD but is enabled separately (and later).

### Related Issues

For https://github.com/Expensify/Expensify/issues/661712

## Deployment

### DIFF
--- a/bin/BedrockWorkerManager.php
+++ b/bin/BedrockWorkerManager.php
@@ -15,6 +15,7 @@ use Expensify\Bedrock\Exceptions\Jobs\IllegalAction;
 use Expensify\Bedrock\Exceptions\Jobs\RetryableException;
 use Expensify\Bedrock\Jobs;
 use Expensify\Bedrock\LocalDB;
+use Expensify\Bedrock\Status;
 
 /*
  * BedrockWorkerManager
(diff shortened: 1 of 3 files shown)


chiragsalian:
Closing — pivoting away from BWM-side pacing to a server-side fix that caps GetOnyxUpdatesForAsyncNotification thread fanout directly in Auth (see Expensify/Auth#23365).

PR #267, opened 2026-08-04
Do not open the circuit breaker when only one host failed

justinpersaud:
# Explanation of Change

The circuit breaker opened when a single node was down. We don't need the breaker to jump in here, only when many hosts are down or the cluster isn't responding.

### Related Issues

For https://github.com/Expensify/Bedrock-PHP/pull/258

For https://github.com/Expensify/Expensify/issues/657702

## Tests

AI tests below

---

No automated suite in this repo, so I ran a manual script in the Expensidev VM (dead loopback ports for unreachable nodes, local Bedrock on 8888 for healthy ones). 6/6:

```
 [PASS] 1: one failed host never opens it, failover still serves
 [PASS] 2: two failed hosts never open it
 [PASS] 3: three failed hosts open it at the threshold
 [PASS] 4: a dead cluster opens it at the threshold
 [PASS] 5: one dead host IS the whole cluster, so it opens
 [PASS] 6: a healthy call clears both keys
```

Phase 1 is the 2026-08-04 shape and fails on `main`. Happy to attach the script.

## Deployment

Version bump and tag still to come.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -958,8 +958,8 @@ class Client implements LoggerAwareInterface
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
-        // A single failed host is already handled by blacklisting it, so it must not trip the whole cluster.
-        if (count($this->failedHosts) < min(2, count($this->mainHostConfigs + $this->failoverHostConfigs))) {
+        // Blacklisting handles fewer, and the cluster only loses its leader once 3 nodes are unreachable.
+        if (count($this->failedHosts) < min(3, count($this->mainHostConfigs + $this->failoverHostConfigs))) {
             return;
         }
         $ttl = $this->circuitBreakerCooldown + 60;
--- a/src/Client.php
+++ b/src/Client.php
@@ -362,6 +362,10 @@ class Client implements LoggerAwareInterface
         }
         try {
             $response = $this->doCall($method, $headers, $body);
+        } catch (TimeoutError $e) {
+            // A timeout is never retried on another host, so it can never fail against all of them, but it still counts.
+            $this->recordCircuitFailure(true);
+            throw $e;
         } catch (BedrockError $e) {
             $this->recordCircuitFailure();
             throw $e;
@@ -950,13 +954,13 @@ class Client implements LoggerAwareInterface
      * the breaker never closes early) and the '-fails' count survives a burst, but short enough to
      * self-clean once traffic goes quiet.
      */
-    private function recordCircuitFailure(): void
+    private function recordCircuitFailure(bool $skipHostCheck = false): void
     {
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
-        // A cluster only loses its leader once 3 nodes are unreachable; below that, blacklisting routes around them.
-        if (count($this->failedHosts) < min(3, count($this->mainHostConfigs + $this->failoverHostConfigs))) {
+        // While any host can still serve the call, blacklisting the bad ones is enough.
+        if (!$skipHostCheck && count($this->failedHosts) < count($this->failoverHostConfigs ?: $this->mainHostConfigs)) {
             return;
         }
         $ttl = $this->circuitBreakerCooldown + 60;
--- a/src/Client.php
+++ b/src/Client.php
@@ -960,7 +960,7 @@ class Client implements LoggerAwareInterface
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
-        // The blacklist already steers us away from the hosts that failed, so while any host is left we don't count it here.
+        // Only count this failure toward the circuit-breaker threshold once every host is in the blacklist. Until then, the blacklist will route future requests to a healthy host, so the cluster as a whole isn't down yet.
         if (!$skipHostCheck && count($this->failedHosts) < count($this->failoverHostConfigs ?: $this->mainHostConfigs)) {
             return;
         }
--- a/src/Client.php
+++ b/src/Client.php
@@ -363,7 +363,7 @@ class Client implements LoggerAwareInterface
         try {
             $response = $this->doCall($method, $headers, $body);
         } catch (TimeoutError $e) {
-            // A timeout is never retried on another host, so it can never fail against all of them, but it still counts.
+            // A timeout doesn't mark the host as failed, so it records none and could never pass the check below.
             $this->recordCircuitFailure(true);
             throw $e;
         } catch (BedrockError $e) {
--- a/src/Client.php
+++ b/src/Client.php
@@ -363,7 +363,7 @@ class Client implements LoggerAwareInterface
         try {
             $response = $this->doCall($method, $headers, $body);
         } catch (TimeoutError $e) {
-            // When a timeout happens, we don't retry the request, so we don't mark the host as failed. 
+            // When a timeout happens, we don't retry the request, so we don't mark the host as failed.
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -443,6 +448,7 @@ private function doCall($method, $headers = [], $body = '')
         $rawRequest .= $body;
 
         $response = null;
+        $this->failedHosts = [];
```
iwiznia, line 456: Probably better in `getPossibleHosts`?

src/Client.php
```diff
@@ -948,6 +955,10 @@ private function recordCircuitFailure(): void
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
+        // A cluster only loses its leader once 3 nodes are unreachable; below that, blacklisting routes around them.
+        if (count($this->failedHosts) < min(3, count($this->mainHostConfigs + $this->failoverHostConfigs))) {
```
iwiznia: Why are you adding main + failovers? That count will be 8 in production.. I think you only want failoverHostConfigs here.

src/Client.php
```diff
@@ -948,6 +955,10 @@ private function recordCircuitFailure(): void
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
+        // A cluster only loses its leader once 3 nodes are unreachable; below that, blacklisting routes around them.
```
iwiznia: Not sure I get this, we can have a cluster with 3 nodes and have 6 in total, so with 3 nodes down (and 3 up) we have a working cluster. 
Shouldn't we be circuit breaking when all hosts are down?

iwiznia (changes requested):
Have to be honest, I am not sure what the circuit breaker is doing or is supposed to be doing... 
Like if one command is bad and gets attempted in all hosts and fails, it will stop processing all requests?

src/Client.php
justinpersaud: This is an array union, so it doesn't add both together. It should still be 6.

src/Client.php
justinpersaud: fair, i will move it to all hosts

justinpersaud:
>I am not sure what the circuit breaker is doing or is supposed to be doing...

The goal of the circuit breaker is to stop creating new requests up to the readTimeout when the auth cluster cannot handle any additional requests due to whatever reason. It prevents overwhelming an overloaded cluster during fires. It is not one bad command, but rather, 10 failures. Any success within that clears it.

src/Client.php
justinpersaud: Why is that? Happy to move but not sure why

src/Client.php
iwiznia: Oh, TIL that's how it works

iwiznia (dismissed):

justinpersaud:
We have a tag conflict here now from another PR by @Valforte. I will have to move this over to 2.3.7.. and then we skip 2.3.6 I guess?

justinpersaud:
actually since this one was already here first @Valforte I am going to use the 2.3.6 tag here and you can use 2.3.7 on yours when it is ready

iwiznia (dismissed):

iwiznia:
ok, make sure that the tag points to the correct commit when you create it.

iwiznia:
Oh you already created it and it's pointing to the correct commit...

justinpersaud:
Yeah we should really just add the tag automation to this repo

src/Client.php
```diff
@@ -357,6 +362,10 @@ public function call($method, $headers = [], $body = '')
         }
         try {
             $response = $this->doCall($method, $headers, $body);
+        } catch (TimeoutError $e) {
+            // A timeout doesn't mark the host as failed, so it records none and could never pass the check below.
```
flodnv: What? This seems like an AI comment that taken out of its context, is meaningless (and very confusing)

src/Client.php
```diff
@@ -943,11 +954,15 @@ private function circuitBreakerAllowsRequest(): bool
      * the breaker never closes early) and the '-fails' count survives a burst, but short enough to
      * self-clean once traffic goes quiet.
      */
-    private function recordCircuitFailure(): void
+    private function recordCircuitFailure(bool $skipHostCheck = false): void
     {
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
+        // While any host can still serve the call, blacklisting the bad ones is enough.
```
flodnv: I also don't understand this comment. Why is it talking about blacklisting, when what we're doing is an early return?

flodnv (commented):
The comments are more confusing than the code 😅

src/Client.php
justinpersaud: Blacklisting prevents requests to a host that have failed for some reason already, but the early return here is because there are still candidate hosts that can serve the request and we don't want to prevent a request from going to them

src/Client.php
justinpersaud: rewording it

src/Client.php
```diff
@@ -357,6 +362,11 @@ public function call($method, $headers = [], $body = '')
         }
         try {
             $response = $this->doCall($method, $headers, $body);
+        } catch (TimeoutError $e) {
+            // When a timeout happens, we no longer mark the host as failed,
+            // but if all hosts are timing out, for the sake of circuit breaking, we want to keep track of this
```
flodnv:
```suggestion
            // When a timeout happens, we don't retry the request, so we don't mark the host as failed. 
            // However, if all hosts are timing out, we need to trip the circuit breaker.
```

Is this correct?

src/Client.php
```diff
@@ -943,11 +955,15 @@ private function circuitBreakerAllowsRequest(): bool
      * the breaker never closes early) and the '-fails' count survives a burst, but short enough to
      * self-clean once traffic goes quiet.
      */
-    private function recordCircuitFailure(): void
+    private function recordCircuitFailure(bool $skipHostCheck = false): void
     {
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
+        // The blacklist already steers us away from the hosts that failed, so while any host is left we don't count it here.
```
flodnv:
```suggestion
        // Only count this failure toward the circuit-breaker threshold once every host is in the blacklist. Until then, the blacklist will route future requests to a healthy host, so the cluster as a whole isn't down yet.
```

flodnv (commented):
To be honest, I don't think I fully understand this PR. Does it mean that if we get 10 different command timeouts in 1 minute, the circuit breaker will kick in?  So if a user causes 10 timeouts in 1 minute, the circuit breaker will kick in and prevent all other requests? If so, that feels off 🤔

justinpersaud:
>Does it mean that if we get 10 different command timeouts in 1 minute, the circuit breaker will kick in? So if a user causes 10 timeouts in 1 minute, the circuit breaker will kick in and prevent all other requests?

Yes, but this behaviour was introduced in the other PR Ionatan made regarding timeouts. Also, this assumes zero requests succeed in that time period. Remember, any successful request resets the counter.

iwiznia (approved):

justinpersaud:
We may not need this PR after all. I misdiagnosed the original issue I was looking at earlier in the week. Details here https://expensify.slack.com/archives/C05CBC62HGW/p1786125723436259?thread_ts=1783698356.886189&cid=C05CBC62HGW

If there are no issues over the weekend with the LAX maintenance, then I think we just close this.

justinpersaud:
No issues, so closing this as not needed

justinpersaud:
(noting I will delete the tags now too)

PR #268, opened 2026-08-06
[HOLD] Retry socket_select() when interrupted by a signal (EINTR)

Valforte:
HOLD while we monitor the original issue to check if this PR is really necessary

# Explanation of Change

During a non-blocking connect, `socket_connect()` returns `EINPROGRESS` (115) and we call `socket_select()` to wait for the socket to become writable. That wait can be interrupted by a signal (`EINTR`), which POSIX expects callers to retry — it says nothing about the health of the host.

`sendRawRequest()` treated any `false` return from `socket_select()` as fatal, which caused three problems on every occurrence:

1. A healthy Bedrock host got `markHostAsFailed()`'d and the request failed over to another host.
2. The un-suppressed `socket_select()` warning leaked a raw PHP Warning into the logs, tripping deploy-blocker detection.
3. The `ConnectionFailure` message read `socket_last_error($this->socket)`, which still holds the stale `EINPROGRESS`. That is why the logs read `Error: Operation now in progress` for what was actually `EINTR`. `socket_select()` reports its failure through the global socket error, so the message now reads that instead.

This PR retries the select when, and only when, it was interrupted by a signal. Retries are bounded by a wall-clock deadline computed once from `connectionTimeout`, so an EINTR storm cannot extend the effective connection timeout. Once the deadline passes, the select returns `0` immediately and the existing timeout `ConnectionFailure` is thrown as before. Every other failure mode keeps its current behavior.

Bumps the version to `2.3.6`.

### Related Issues

$ https://github.com/Expensify/Expensify/issues/665789

## Deployment

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "expensify/bedrock-php",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "2.3.5",
+    "version": "2.3.6",
     "authors": [
         {
             "name": "Expensify",
--- a/src/Client.php
+++ b/src/Client.php
@@ -608,14 +608,22 @@ class Client implements LoggerAwareInterface
             if ($socketErrorCode === 115) {
                 $this->logger->debug('Bedrock\Client - socket_connect returned error 115, waiting for connection to complete.', ['host' => $host]);
 
-                // Wait for the socket to be ready for writing after EINPROGRESS
-                $write = [$this->socket];
-                $read = [];
-                $except = [];
-                $selectResult = socket_select($read, $write, $except, $this->connectionTimeout, $this->connectionTimeoutMicroseconds);
+                // Wait for the socket to be ready for writing after EINPROGRESS, retrying if a signal interrupts the
+                // wait (EINTR), which says nothing about the health of the host. The deadline bounds those retries.
+                $deadline = microtime(true) + $this->connectionTimeout + $this->connectionTimeoutMicroseconds / 1000000;
+                do {
+                    $write = [$this->socket];
+                    $read = [];
+                    $except = [];
+                    $secondsLeft = max(0, $deadline - microtime(true));
+                    $selectResult = @socket_select($read, $write, $except, (int) $secondsLeft, (int) (($secondsLeft - (int) $secondsLeft) * 1000000));
+
+                    // socket_select() sets the global socket error, not the error of any individual socket.
+                    $wasInterrupted = $selectResult === false && socket_last_error() === SOCKET_EINTR;
+                } while ($wasInterrupted && $secondsLeft > 0);
 
                 if ($selectResult === false) {
-                    $socketError = socket_strerror(socket_last_error($this->socket));
+                    $socketError = socket_strerror(socket_last_error());
                     throw new ConnectionFailure("socket_select failed after EINPROGRESS for $host:$port. Error: $socketError");
                 } elseif ($selectResult === 0) {
                     throw new ConnectionFailure("Socket not ready for writing within timeout after EINPROGRESS for $host:$port");


justinpersaud:
Hey @Valforte, did you do any testing here to confirm this works? I'm wondering why the connect gets interrupted?

Valforte:
## Testing

Verified inside the Expensidev VM (Linux, PHP 8.3.25) — the platform where this branch matters, since `Client.php` gates the wait on EINPROGRESS being errno `115`.

### 1. The condition the fix branches on

`socket_select()` reports its own failure through the **global** socket error, not through the error of the socket passed to it:

```
after connect:  perSocket=115 global=115
select returned: false
after select:   perSocket=115 global=4      <- EINTR
global == SOCKET_EINTR: true
perSocket == SOCKET_EINTR: false
```

This is why the check reads argless `socket_last_error()`. It is also why the old exception message said `Error: Operation now in progress` for what was really `EINTR` — it read back the stale per-socket `EINPROGRESS`, exactly as seen in the production logs on the linked issue.

### 2. Old vs new, isolated

A child process sends `SIGUSR1` every 3ms while the parent waits on a connect to `8.8.8.8:4444`, which drops the SYN so the connect never completes and `socket_select()` blocks. `connectionTimeout` is 2s.

| | selects | EINTRs absorbed | elapsed | outcome |
|---|---|---|---|---|
| before | 1 | 0 | 0.051s | throws `socket_select failed after EINPROGRESS` — healthy host blacklisted, failover |
| after | 39 | 38 | 2.003s | `Socket not ready for writing within timeout` — the honest error |

### 3. Through the real `Client`

Same signal storm, driving `Client::call('Ping')`, with an error handler that mirrors Web-Expensify's `_error_handler` (returns early when `error_reporting()` equals the `@`-suppression mask):

```
using: /vagrant/Bedrock-PHP/src/Client.php
contains retry loop: YES
elapsed=2.007s (connectionTimeout=2s) unsuppressedWarnings=0
outcome: ConnectionFailure: Socket not ready for writing within timeout after EINPROGRESS for 8.8.8.8:4444
```

`unsuppressedWarnings=0` confirms the `@` stops the PHP Warning from reaching the handler, so it no longer creates a ServerError/deploy blocker. `_errorUtilities.php:162` is the early return that makes this work.

### 4. Retries stay inside the timeout

`2.003s` and `2.007s` against a 2s `connectionTimeout`: the deadline is computed once before the loop, so an EINTR storm cannot extend the effective connect timeout no matter how many signals arrive.

### 5. Connect still succeeds after being interrupted

Pointed at a host that accepts (`8.8.8.8:53`) with signals every 3ms: 6 selects, 5 EINTRs absorbed, `CONNECTED` in 0.019s. Retrying does not break the normal path. This one was run standalone rather than through `Client`.

### Note for anyone verifying locally

The VM sets `auto_prepend_file => /git/switch/_beforeSwitch.php`, which eagerly loads Web-Expensify's vendored copy of this library. `Expensify\Bedrock\Client` is therefore already declared before any test script runs, so a script under `Bedrock-PHP/` silently exercises `/vagrant/Web-Expensify/vendor/expensify/bedrock-php/src/Client.php` — the released version, not your branch. Run with `php -d auto_prepend_file= ...`, and assert on `(new ReflectionClass(Client::class))->getFileName()` before trusting any result.

Valforte:
@justinpersaud I honestly don't know how I could properly test this change, so I asked claude for help testing it and it came up with a script I could run to validate it before and after the changes. I'm having it post a comment with the testing validations. If you still don't feel confortable merging the change, I understand, as this is a very delicate part of the code. The error also happened only on two occasions that are probably related to server issues (I'll check, but the first time I checked I couldn't find anything definitive). We could leave it as-is and I keep monitoring the bugbot for recurrence for a few more weeks if you think it's best.

Valforte:
This is the file claude create to test:

`./script/client-eintr-test.php`
```php
<?php
/**
 * Drives the real Expensify\Bedrock\Client through an EINTR storm.
 * Must run on Linux: Client.php checks for EINPROGRESS as errno 115, which is Linux-specific.
 *
 * Usage (inside the VM):
 *   cd /vagrant/Bedrock-PHP && composer install && php script/client-eintr-test.php
 */

require __DIR__.'/../vendor/autoload.php';

// The VM registers Web-Expensify's Composer autoloader, which maps Expensify\Bedrock\ to its own
// installed copy of this library. Prepend an autoloader so this checkout's src wins instead.
spl_autoload_register(function ($class) {
    if (str_starts_with($class, 'Expensify\\Bedrock\\')) {
        $path = __DIR__.'/../src/'.str_replace('\\', '/', substr($class, strlen('Expensify\\Bedrock\\'))).'.php';
        if (file_exists($path)) {
            require $path;
        }
    }
}, true, true);

use Expensify\Bedrock\Client;
use Expensify\Bedrock\Exceptions\BedrockError;

if (PHP_OS_FAMILY !== 'Linux') {
    exit("EINPROGRESS is errno 115 on Linux only; Client.php will not take that branch here.\n");
}

$loadedFrom = (new ReflectionClass(Client::class))->getFileName();
echo "using: $loadedFrom\n";
echo 'contains retry loop: '.(str_contains(file_get_contents($loadedFrom), '$wasInterrupted') ? "YES\n" : "NO\n");

$warnings = 0;
set_error_handler(function ($errno, $errstr) use (&$warnings) {
    // Mirror Web-Expensify's handler: honour the @ suppression operator.
    if (error_reporting() === (E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE)) {
        return true;
    }
    if (str_contains($errstr, 'socket_select')) {
        $warnings++;
        echo "  [unsuppressed php warning] $errstr\n";
    }
    return true;
});

// Signal storm from a child so that socket_select() is interrupted repeatedly.
pcntl_async_signals(true);
pcntl_signal(SIGUSR1, function () {});
$child = pcntl_fork();
if ($child === 0) {
    $parent = posix_getppid();
    while (posix_kill($parent, SIGUSR1)) {
        usleep(3000);
    }
    exit(0);
}

// 8.8.8.8:4444 drops the SYN, so connect() stays in progress and socket_select() blocks.
Client::configure([
    'clusterName' => 'eintr-test',
    'mainHostConfigs' => ['8.8.8.8' => ['blacklistedUntil' => 0, 'port' => 4444]],
    'failoverHostConfigs' => [],
    'connectionTimeout' => 2,
    'connectionTimeoutMicroseconds' => 0,
    'circuitBreakerThreshold' => 0,
]);

$start = microtime(true);
try {
    Client::getInstance()->call('Ping');
    $outcome = 'unexpectedly succeeded';
} catch (BedrockError $e) {
    $outcome = get_class($e).': '.$e->getMessage();
}
$elapsed = round(microtime(true) - $start, 3);

echo "elapsed={$elapsed}s (connectionTimeout=2s) unsuppressedWarnings=$warnings\n";
echo "outcome: $outcome\n\n";
echo "PASS looks like: elapsed ~2s, 0 unsuppressed warnings, message 'Socket not ready for writing within timeout'.\n";
echo "The old code fails fast (<0.2s) with 'socket_select failed after EINPROGRESS' and 1+ warnings.\n";

posix_kill($child, SIGKILL);
pcntl_waitpid($child, $status);
```

Run with:
```
php -d auto_prepend_file= script/client-eintr-test.php
```
from the VM in `main` to emulate the bugbot issue or in `valforte-retry-socket-select-eintr` to check the fix. Here are the outputs 

`main`
```
vagrant@expensidev-2404  /vagrant/Bedrock-PHP  ↰ main  php -d auto_prepend_file= script/client-eintr-test.php 
using: /vagrant/Bedrock-PHP/src/Client.php
contains retry loop: NO
  [unsuppressed php warning] socket_select(): Unable to select [4]: Interrupted system call
elapsed=0.005s (connectionTimeout=2s) unsuppressedWarnings=1
outcome: Expensify\Bedrock\Exceptions\ConnectionFailure: socket_select failed after EINPROGRESS for 8.8.8.8:4444. Error: Operation now in progress

PASS looks like: elapsed ~2s, 0 unsuppressed warnings, message 'Socket not ready for writing within timeout'.
The old code fails fast (<0.2s) with 'socket_select failed after EINPROGRESS' and 1+ warnings.
```

`valforte-retry-socket-select-eintr`
```
vagrant@expensidev-2404  /vagrant/Bedrock-PHP   valforte-retry-socket-select-eintr  php -d auto_prepend_file= script/client-eintr-test.php 
using: /vagrant/Bedrock-PHP/src/Client.php
contains retry loop: YES
elapsed=2.012s (connectionTimeout=2s) unsuppressedWarnings=0
outcome: Expensify\Bedrock\Exceptions\ConnectionFailure: Socket not ready for writing within timeout after EINPROGRESS for 8.8.8.8:4444

PASS looks like: elapsed ~2s, 0 unsuppressed warnings, message 'Socket not ready for writing within timeout'.
The old code fails fast (<0.2s) with 'socket_select failed after EINPROGRESS' and 1+ warnings.
```

justinpersaud:
I think the Claude test is fair. I do that too, it's just nice to have that validation in here before we touch this code.

We've played around with this code a lot in the past.

@MelvinBot can you elaborate on this PR and https://github.com/Expensify/Expensify/issues/665789?

What is causing the socket_select to be interrupted by a signal in the first place? Are you sure that is happening, and if so, provide evidence from logs, do not make assumptions.

Valforte:
I don't believe @MelvinBot has access to this repo

justinpersaud:
Ok I did some analysis on my own to understand this a bit better and it appears the connection can be interrupted when we do a php-fpm reload, but we should figure out why that is happening. I'm not sure if this is the right solution and won't have time to get back to it before going OOO, so I am going to reassign this over to @flodnv who I have worked with on this code previously.

flodnv:
> we should figure out why that is happening. I'm not sure if this is the right solution

Agree. I closed the issue to see if it happens again (it hasn't happened in 10 days).

Valforte:
Didn't happen again for a long time, closing

PR #270, opened 2026-08-13
Add unique-as-retry support to Bedrock-PHP and BWM

inimaga:
# Explanation of Change

Extends the Jobs client with uniqueAsRetry and enqueueVersion parameters, then updates BWM to return each job’s starting version when finishing or retrying it. This allows Bedrock to detect stale worker results and preserve newer enqueue data.

### Related Issues

https://github.com/Expensify/Expensify/issues/670760

https://github.com/Expensify/Bedrock/pull/2723

## Deployment

- [x] I followed the steps in the [README](../blob/main/README.md#publishing-your-changes) to ensure this PR is deployed properly

PR #273, opened 2026-08-21
Forward job data snapshots for reruns

inimaga:
### Details

Following the merged [Bedrock #2723](https://github.com/Expensify/Bedrock/pull/2723) PR, this PR updates the PHP client and worker manager to use `rerunIfDataChanged` and send the original job data when finishing, retrying, or failing a job. Bedrock can then preserve newer work and run it again. 

The old version-tracking fields introduced via the PR https://github.com/Expensify/Bedrock-PHP/pull/270 are removed, and snapshots preserve JSON objects and arrays to avoid unnecessary reruns.

### Fixed/Related Issue

https://github.com/Expensify/Expensify/issues/670760

## Deployment


inimaga:
On hold for now.

src/Jobs.php
```diff
@@ -170,27 +172,48 @@ public function call($method, $headers = [], $body = '')
             throw new GenericError("Generic error for job $job");
         }
 
+        if (($method === 'GetJob' && $response['body'] !== []) || $method === 'GetJobs') {
```
iwiznia, line 175: I am not sure I get this (and the handling of expectedData [here](https://github.com/Expensify/Bedrock-PHP/pull/273/changes#diff-f5e368bd2af36b1fb61a893954ab2deb00183ef0bf6bd2c43659a87d3314da9fR137) too), why do we need to decode and reencode the data if PHP does not use it for anything other than passing it back as is? Maybe we should've made bedrock return the data as string so it never gets decoded/encoded and avoid all this.

iwiznia (changes requested):

src/Jobs.php
inimaga: > Maybe we should've made bedrock return the data as string so it never gets decoded/encoded and avoid all this.

Yeah... Agreed.

inimaga: https://github.com/Expensify/Bedrock/pull/2803

pecanoro: I checked the Bedrock PR and I also find it odd. Why do we need to do all this encoding and decoding instead of passing back what we get from GetJob directly? Isn't it a JSON object already?

src/Jobs.php
```diff
@@ -177,20 +177,20 @@ public function call($method, $headers = [], $body = '')
...
+     * @param string|null $connection         (optional) Specify 'Connection' header using constants defined in this class.
+     * @param string|null $retryAfter         (optional) Specify after what time in RUNNING this job should be retried (same syntax as repeat)
+     * @param bool        $overwrite          (optional) Only applicable when unique is is true. When set to true it will overwrite the existing job with the new jobs data
+     * @param bool        $rerunIfDataChanged (optional) Requeue when an enqueue changes a running job's data. Requires unique and overwrite; cannot own child jobs.
      *
      * @return array Containing "jobID"
      */
-    public function createJob($name, $data = null, $firstRun = null, $repeat = null, $unique = false, $priority = self::PRIORITY_MEDIUM, $parentJobID = null, $connection = self::CONNECTION_WAIT, $retryAfter = null, $overwrite = true, $uniqueAsRetry = false)
+    public function createJob($name, $data = null, $firstRun = null, $repeat = null, $unique = false, $priority = self::PRIORITY_MEDIUM, $parentJobID = null, $connection = self::CONNECTION_WAIT, $retryAfter = null, $overwrite = true, $rerunIfDataChanged = false)
```
iwiznia:
```suggestion
    public function createJob($name, $data = null, $firstRun = null, $repeat = null, $unique = false, $priority = self::PRIORITY_MEDIUM, $parentJobID = null, $connection = self::CONNECTION_WAIT, $retryAfter = null, $overwrite = true, bool $rerunIfDataChanged = false)
```

iwiznia (changes requested):
This needs to hold on the bedrock PR, no?

bin/BedrockWorkerManager.php
```diff
@@ -322,7 +322,8 @@
                 $job['name'] = $jobParts[0];
                 $workerName = explode('/', $job['name'])[1];
                 $workerFilename = $workerPath."/$workerName.php";
-                $enqueueVersion = isset($job['enqueueVersion']) ? (int) $job['enqueueVersion'] : null;
+                // Preserve the dequeued data so Bedrock can detect changes made while the worker is running.
```
pecanoro, line 326:
```suggestion

                // Preserve the dequeued data so Bedrock can detect changes made while the worker is running.
```

src/Jobs.php
pecanoro: I would also add bool to overwrite since it is a new param

src/Jobs.php
```diff
@@ -323,21 +323,21 @@ public function updateJob($jobID, $data, $repeat = null, ?int $priority = null,
     /**
      * Marks a job as finished, which causes it to repeat if requested.
      *
-     * @param int      $jobID
-     * @param array    $data           (optional)
-     * @param int|null $enqueueVersion (optional) Version returned when the job was dequeued
+     * @param int         $jobID
+     * @param array       $data         (optional)
```
pecanoro:
```suggestion
     * @param array|null       $data         (optional)
```

pecanoro (changes requested):

inimaga:
Actually, this doesn't need to be held as `rerunIfDataChanged` hasn't been enabled yet downstream.

inimaga:
@pecanoro bump for review as @iwiznia is out of office

pecanoro (approved):

pecanoro:
Discussing the other PR with Issa before merging this one!

PR #274, opened 2026-08-24
Tag new versions automatically when a pull request merges

justinpersaud:
# Explanation of Change

Automating version tag bumps which has been tedious and error prone

### Related Issues

N/A


composer.json
```diff
@@ -2,7 +2,6 @@
     "name": "expensify/bedrock-php",
     "description": "Bedrock PHP Library",
     "type": "library",
-    "version": "2.3.7",
```
iwiznia, line 5: I don't think this is ok, I thought a package has to have a version for it to work properly... php-libs has one for example

.github/workflows/version.yml
```diff
@@ -0,0 +1,84 @@
...
+        if [ -z "$LATEST_VERSION" ]; then
+          echo "Found no version tag to increment."
+          exit 1
+        fi
+        MAJOR="${LATEST_VERSION%%.*}"
+        MINOR="${LATEST_VERSION#*.}"
+        MINOR="${MINOR%%.*}"
+        PATCH="${LATEST_VERSION##*.}"
+        echo "Latest version: $LATEST_VERSION"
+        echo "NEW_VERSION=$MAJOR.$MINOR.$((10#$PATCH + 1))" >> "$GITHUB_ENV"
```
iwiznia, line 58: So we never increment neither minor nor major? Should we just then have major and increment that?

justinpersaud: Yeah we only increment patch version here, unless we add a way to increment minor/major too automatically. You can do that manually today, and then this will just continue from there.

e.g. if you push 3.0.0, the next time this runs it will just do 3.0.1

Do you want me to add a way to do minor/major here too?

composer.json
justinpersaud: The docs say to omit this if the repository is using tagged versions: https://getcomposer.org/doc/04-schema.md#version

.github/workflows/version.yml
iwiznia: I think having major, minor and patch is useless if they mean nothing, so maybe let's just have major versions?
NAB though

composer.json
iwiznia: Oh TIL

iwiznia (approved):
I assume this will be self tested when we merge this right?
Given that, will let you hit merge, in case it fails you can debug/fix.

justinpersaud:
Yeah I can test it live, but it won't work when this one is merged because the action file isn't there yet. It would work on the next PR after.

justinpersaud:
oh so it does run on this PR, interesting

PR #275, opened 2026-08-24
Give writes and reads separate circuit breakers

justinpersaud:
# Explanation of Change

Split up the circuit breaker by different scopes, mainly read and write. There are some other cases below, where we send a request and the caller doesn't specify. See below.

### Scopes

| Key | Set by |
| --- | --- |
| `auth\|write`, `auth\|read` | Web-Expensify, via the existing `GetWriteCommands` map |
| `auth\|unknown` | Web-Expensify, when that map's refresh fails — it is itself an Auth call |
| `auth\|default` | Web-Secure (`Web-Secure/lib/Auth.php:426`) |
| `webrock\|default` | `Jobs` queue operations, `Cache`, `DB`, `Status`, `ChatBot\Bedrock` |

## Tests

Claude wrote a test script to run locally, adn then ran test cases below. I can share the code if you care to look at it.

```
 [PASS] 1. 9 consecutive write failures stays closed
 [PASS] 1. the 10th failure opens the write scope
 [PASS] 1. the read scope is unaffected
 [PASS] 2. 200 read successes do not clear the write trip
 [PASS] 3. a write success resets the run, so 9 + 1 + 9 stays closed
 [PASS] 4. a sustained 90% failure rate with no run of 10 stays closed, as designed
 [PASS] 5. threshold 0 disables the breaker entirely
 [PASS] 6. the scope is open immediately after tripping
 [PASS] 6. the scope closes itself once the cooldown expires
 [PASS] 7. a zero cooldown throws at construction
 [PASS] 8. real failures through call() trip the write scope then fail fast (10 attempted, 5 rejected)
 [PASS] 8. the fail-fast message names the scope
 [PASS] 8. read calls are still admitted while the write scope is open
```

### Related Issues

For https://github.com/Expensify/Expensify/issues/673586

## Deployment

- [x] Once this merges and a version is tagged, I will bump the `expensify/bedrock-php` constraint in Web-Expensify and Web-Secure if they need this change.


iwiznia:
> auth\|default | Web-Secure (Web-Secure/lib/Auth.php:426)

Why secure is not using the same as web?

> webrock\|default | Jobs queue operations, Cache, DB, Status, ChatBot\Bedrock

I assume no distinguishing between read and write since we don't have getWriteCommands in webrock, right?

src/Client.php
```diff
@@ -329,31 +330,40 @@ public function getClusterName(): string
...
+     *
+     * @param string      $method       Request method
+     * @param array       $headers      Request headers (optional)
+     * @param string      $body         Request body (optional)
+     * @param string|null $breakerScope Scope to account this call against, 'default' when not given
      *
      * @return array JSON response
      */
-    public function call($method, $headers = [], $body = '')
+    public function call($method, $headers = [], $body = '', ?string $breakerScope = null)
```
iwiznia: Should we add the scope to `headers` since we pass everything there?

src/Client.php
```diff
@@ -329,31 +330,40 @@ public function getClusterName(): string
...
     {
-        if (!$this->circuitBreakerAllowsRequest()) {
-            $this->logger->info('Bedrock\Client - Circuit breaker open, failing fast', ['clusterName' => $this->clusterName]);
-            throw new ConnectionFailure("Bedrock circuit breaker open for cluster $this->clusterName");
+        // The argument is optional so existing callers keep working, and the key needs a scope segment
+        // either way, so everything that does not classify its calls shares this one breaker.
+        $scope = $breakerScope ?? 'default';
```
iwiznia: You should remove this and change `?string $breakerScope = null` to `string $breakerScope = 'default'`

src/Client.php
```diff
@@ -904,69 +914,64 @@ private function isApcuAvailable(): bool
...
 
     /**
-     * Returns false when the per-cluster circuit breaker is open, so the caller fails fast instead of
+     * Builds the APCu key for one piece of a scope's breaker state: 'fails' holds the current run of
+     * failures, 'open' is the trip marker whose presence means the breaker is open. '|' cannot appear in
+     * a cluster or scope name, so no two scope/suffix pairs can produce the same key.
+     */
+    private function breakerKey(string $scope, string $suffix): string
+    {
+        return self::CIRCUIT_BREAKER_CACHE_PREFIX."$this->clusterName|$scope|$suffix";
```
iwiznia: You changed the key from `-` to `|` on purpose?

src/Client.php
```diff
@@ -904,69 +914,64 @@ private function isApcuAvailable(): bool
...
+    private function circuitBreakerAllowsRequest(string $scope): bool
     {
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return true;
         }
-        $openUntil = apcu_fetch(self::CIRCUIT_BREAKER_CACHE_PREFIX.$this->clusterName.'-open');
 
-        return $openUntil === false || $openUntil <= time();
+        // The marker's TTL is the cooldown, so its presence alone means the breaker is still open.
+        return !apcu_exists($this->breakerKey($scope, 'open'));
```
iwiznia, line 937: 😕 why did you remove the time check?

src/Client.php
```diff
@@ -904,69 +914,64 @@ private function isApcuAvailable(): bool
...
         if ($this->circuitBreakerThreshold <= 0 || !$this->isApcuAvailable()) {
             return;
         }
-        $failsKey = self::CIRCUIT_BREAKER_CACHE_PREFIX.$this->clusterName.'-fails';
-        $openKey = self::CIRCUIT_BREAKER_CACHE_PREFIX.$this->clusterName.'-open';
         // Only take the write lock when there is actually state to clear, so a successful call on a
         // healthy cluster (the vast majority) does cheap reads and no writes.
-        if (apcu_fetch($failsKey) !== false || apcu_fetch($openKey) !== false) {
+        $failsKey = $this->breakerKey($scope, 'fails');
+        if (apcu_exists($failsKey)) {
```
iwiznia, line 973: Oh is this why you removed the time check above?

iwiznia (changes requested):

src/Client.php
justinpersaud: hmm no there is no good reason actually -- I saw claude did this for some other things but missed this. I will have it move it back to `-`

src/Client.php
justinpersaud: responding below

src/Client.php
justinpersaud: yeah it used to be cooldown + 60, but now the time to live is the cooldown. if it exists at all, it is open

justinpersaud:
>Why secure is not using the same as web?

it doesn't have `GetWriteCommands`, plus the amount of bedrock calls that come from secure is barely higher than 0%

>I assume no distinguishing between read and write since we don't have getWriteCommands in webrock, right?

correct

iwiznia:
> it doesn't have GetWriteCommands

You can easily add it

> plus the amount of bedrock calls that come from secure is barely higher than 0%

ok

iwiznia (approved):
Not merging since you added Daniel as reviewer, but FYI he's ooo

justinpersaud:
ah ok will merge and open the web-E PR for it

PR #276, opened 2026-08-25
Fetch tags with fetch-depth 0 so the release can find them

justinpersaud:
# Explanation of Change

The bump version failed because it didn't pull in the latest tags, and apparently with fetch-depth 0, it should fix that. Also adding a way to manually create a new release if needed too.

### Related Issues

N/A

## Deployment

- [ ] Once this merges and a version is tagged, I will bump the `expensify/bedrock-php` constraint in Web-Expensify and Web-Secure if they need this change.


iwiznia (approved):
So weird that `fetch-tags` did not fetch tags 😕 
Same as before, will let you merge to debug it

justinpersaud:
there we go, working now

